PR #5, opened 2015-11-20
feat(builder): add kubectl and run slugbuilder as a pod

smothiki:


.gitignore
```diff
@@ -1,3 +1,4 @@
 rootfs/bin/boot
+rootfs/usr/bin/*
```
mboersma: Isn't this already covered by the `rootfs/usr/bin/` exclusion below?

smothiki: opps My bad will remove it. But even though there is a git ignore endpoint while doing git push its still considering binaries in that direcotry

smothiki:
closing this in favor of  https://github.com/deis/builder/pull/8

arschles:
@smothiki can you cherry pick commits from https://github.com/smothiki/builder/pull/1 into #8

PR #12, opened 2015-12-02
fix(builder): add check to locate Procfile

smothiki:
Previously we are just checking procfile present and falling back to dockerfile and not checking if dockerfile is also present.
The new condition will extract the git archive to temp dir and checks for procfile or dockerfile.

### DIFF
--- a/manifests/deis-builder-service.yaml
+++ b/manifests/deis-builder-service.yaml
@@ -12,6 +12,9 @@ spec:
       targetPort: 2223
       name: ssh
       protocol: TCP
+    - port: 3000
+      name: fetcher
+      protocol: TCP
   selector:
     # Match the selector in the RC
     app: deis-builder
--- a/rootfs/etc/confd/templates/builder
+++ b/rootfs/etc/confd/templates/builder
@@ -64,17 +64,18 @@ TMP_DIR=$(mktemp -d -p $BUILD_DIR)
 
 cd $REPO_DIR
 # use Procfile if provided, otherwise try default process types from ./release
+git archive --format=tar.gz ${GIT_SHA} > ${APP_NAME}.tar.gz
+tar -xzf ${APP_NAME}.tar.gz -C $TMP_DIR/
 USING_DOCKERFILE=true
-if [ -f Procfile ]; then
-    PROCFILE=$(cat Procfile | yaml2json-procfile)
+if [ -f $TMP_DIR/Procfile ]; then
+    PROCFILE=$(cat $TMP_DIR/Procfile | yaml2json-procfile)
     USING_DOCKERFILE=false
-    POD="slugbuilder"
 else
     PROCFILE="{}"
 fi
 
 if [[ ! -f /var/run/secrets/object/store/access-key-id  ]]; then
-  if [ -f Dockerfile ]; then
+  if $USING_DOCKERFILE ; then
     l1=`grep -n "object-store" /etc/deis-dockerbuilder.yaml | head -n1 |cut -d ":" -f1`
     l2=$(($l1+3))
     sed "$l1,$l2 d" /etc/deis-dockerbuilder.yaml > /etc/${SLUG_NAME}.yaml.tmp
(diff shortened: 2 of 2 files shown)


rootfs/etc/confd/templates/builder
```diff
@@ -64,17 +64,30 @@ TMP_DIR=$(mktemp -d -p $BUILD_DIR)
...
 else
     PROCFILE="{}"
 fi
+# if [ -f Procfile ]; then
+#     PROCFILE=$(cat Procfile | yaml2json-procfile)
+#     USING_DOCKERFILE=false
+#     POD="slugbuilder"
+# else
+#     PROCFILE="{}"
+# fi
```
arschles: can you delete this?

rootfs/etc/confd/templates/builder
```diff
@@ -64,17 +64,30 @@ TMP_DIR=$(mktemp -d -p $BUILD_DIR)
 
 cd $REPO_DIR
 # use Procfile if provided, otherwise try default process types from ./release
+
+git archive --format=tar.gz ${GIT_SHA} > ${APP_NAME}.tar.gz
```
bacongobbler: why are we archiving then immediately un-tar'ing the directory elsewhere? Can't we just run `cp -R` instead?

bacongobbler: also from the changes here I don't see any technical reason why we need to retain a second copy of the application files in $TMP_DIR.

rootfs/etc/confd/templates/builder
```diff
@@ -64,17 +64,30 @@ TMP_DIR=$(mktemp -d -p $BUILD_DIR)
...
+#     POD="slugbuilder"
+# else
+#     PROCFILE="{}"
+# fi
+echo "procfile" $PROCFILE
+echo "dockerfile" $USING_DOCKERFILE
 
 if [[ ! -f /var/run/secrets/object/store/access-key-id  ]]; then
-  if [ -f Dockerfile ]; then
+  if $USING_DOCKERFILE ; then
```
bacongobbler: IIRC the reason why we didn't go this route was because assigning envvars a value in if statements are not retained after the clause ends, so this will always end up as true. I might be wrong about that but we should be cautious about this.

smothiki: I did a ls of files inside repo direcotry and it didn't show up any files related to example-go, but when I did archive and untar the files are there. You can play with smothiki/builder:3.1.0 to observe this behavior. reason for doing a git archive and then untar the files in a temp direcotry. 
This might be issue with previous builder as well but its passing because we have procfile check from generated slug.

smothiki: regarding bash variables USING_DOCKERFILE is a global variable which is tru by default and being set as false if there is a procfile. Even though someone sets it as environment variable it might not affect the logic.

rootfs/etc/confd/templates/builder
smothiki: I did a ls of files inside repo direcotry and it didn't show up any files related to example-go, but when I did archive and untar the files are there. You can play with smothiki/builder:3.1.0 to observe this behavior. reason for doing a git archive and then untar the files in a temp direcotry.
This might be issue with previous builder as well but its passing because we have procfile check from generated slug.

bacongobbler: > I did a ls of files inside repo direcotry and it didn't show up any files related to example-go

It'll all be hiding under the build dir, which will be in `/home/git/example-go/build/<some-random-build-id>`. That's where it's extracted for the build, and that's how we can detect if a Dockerfile is present.

bacongobbler: After working offline with @smothiki on this, I wasn't aware that this functionality was [previously removed](https://github.com/deis/builder/commit/ec8b981052c5ef434d5370b5778c3167c173010e#diff-cf0e34247cf4169bf1b23bd6817d6e10L67), so essentially this is just adding it back in.

manifests/deis-builder-service.yaml
```diff
@@ -12,6 +12,10 @@ spec:
       targetPort: 2223
       name: ssh
       protocol: TCP
+    - port: 3000
+      targetPort: 3000
```
bacongobbler: this is redundant, so let's remove this: https://github.com/deis/charts/issues/3

manifests/deis-builder-service.yaml
```diff
@@ -12,6 +12,10 @@ spec:
...
+      name: fetcher
+      protocol: TCP
```
bacongobbler: same here: https://github.com/deis/charts/issues/4

manifests/deis-builder-rc.yaml
```diff
@@ -20,6 +20,7 @@ spec:
       containers:
         - name: deis-builder
           image: quay.io/deisci/builder:v2-alpha
+          # image: smothiki/builder:3.1.3
```
bacongobbler: we should probably remove this comment.

rootfs/etc/confd/templates/builder
```diff
@@ -94,7 +101,7 @@ else
   fi
 fi
 
-git archive --format=tar.gz ${GIT_SHA} > ${APP_NAME}.tar.gz
+
```
bacongobbler: let's remove this extra whitespace

manifests/deis-builder-rc.yaml
mboersma: I'd like this comment removed as well, then I'm LGTM on this one.

PR #9, opened 2015-11-24
feat(*): store slugs in minio & launch new pods to run builds

arschles:
stores code in minio and launches new pods to run [slugbuilder](https://github.com/deis/slugbuilder]. also includes experimental support for launching [dockerbuilder](https://github.com/deis/dockerbuilder) pods.

Fixes https://github.com/deis/builder/issues/7 and replaces https://github.com/smothiki/builder/pull/2/files

Related: https://github.com/deis/slugbuilder/pull/6

PR #15, opened 2015-12-09
fix(Makefile): change VERSION, add and fix targets

arschles:
- adding go test to test target
- removing build target from docker-build
- deleting the service on kube-clean

replaces part of #9

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,6 @@ STANDALONE := extract-types  generate-buildhook yaml2json-procfile
 # Docker Root FS
 BINDIR := ./rootfs
 
-DEV_REGISTRY ?= $(shell docker-machine ip deis):5000
 DEIS_REGISTRY ?= ${DEV_REGISTRY}/
 
 # Kubernetes-specific information for RC, Service, and Image.
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,3 @@
-# Short name: Short name, following [a-zA-Z_], used all over the place.
-# Some uses for short name:
-# - Docker image name
-# - Kubernetes service, rc, pod, secret, volume names
 SHORT_NAME ?= builder
 
 # Enable vendor/ directory support.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -19,8 +15,7 @@ STANDALONE := extract-types  generate-buildhook yaml2json-procfile
 # Docker Root FS
 BINDIR := ./rootfs
 
-# Legacy support for DEV_REGISTRY, plus new support for DEIS_REGISTRY.
-DEV_REGISTRY ?= $$DEV_REGISTRY
+DEV_REGISTRY ?= $(shell docker-machine ip deis):5000
```
bacongobbler: we recently removed this from etcd to get the deployments working again: https://github.com/deis/etcd/pull/19

travis was failing because docker-machine was not found, even though we explicitly added DEIS_REGISTRY.

Makefile
```diff
@@ -50,8 +45,11 @@ build:
 		$(call check-static-binary,$(BINARY_DEST_DIR)/$$i); \
 	done
 
-docker-build: build
-	docker build --rm -t $(IMAGE) rootfs
+test:
+	go test ./pkg && go test ./pkg/confd && go test ./pkg/env && go test ./pkg/etcd && go test ./pkg/git && go test ./pkg/sshd
```
bacongobbler: can we use `go test ./...` here so we don't explicitly have to call them out?

Makefile
```diff
@@ -71,10 +69,8 @@ kube-rc:
 	kubectl create -f ${RC}
 
 kube-clean:
-	kubectl delete rc deis-builder
-
-test:
-	@echo "Implement functional tests in _tests directory"
```
bacongobbler: we should retain this. It shows that we intend to add tests, and that `make test` is the way to run them all.

Makefile
mboersma: or `go test $(glide nv)` to avoid the vendor/ directory?

arschles: unfortunately neither of those will work, because there are multiple `package main`s in `./pkg/src/` (each builds a small binary that builder uses).

I have a larger refactor in mind that would remove all those small binaries, which would have the side effect of being able to run `go test ./...`, but for now I'd like to leave as-is. Thoughts?

Makefile
arschles: should I rename the [above `test` target](https://github.com/deis/builder/pull/15#discussion-diff-47140079R48) then?

arschles:
@bacongobbler removed `DEV_REGISTRY`

Makefile
bacongobbler: oh nvm. Sick me does not have the mental capacity to remember the test target.

PR #18, opened 2015-12-09
feat(*): add support for uploading tarballs to minio

arschles:
replaces part of #9

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -1,5 +1,4 @@
-/*Package etcd is a library for performing common Etcd tasks.
- */
+// Package etcd is a library for performing common Etcd tasks.
 package etcd
 
 import (
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -8,7 +8,6 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


pkg/etcd/etcd.go
```diff
@@ -65,7 +71,19 @@ type GetterSetter interface {
...
+		host = defaultHost
+	}
+	if err != nil {
+		port = 4001
+	}
+
+	// Aaron(12/9/2015): to shame me into fixing this later, but it's a bigger change
+	fmt.Println("Aaron hasn't fixed the env var collection stage in the cookoo route yet...")
+	// url := p.Get("url", "http://localhost:4001").(string)
+	url := fmt.Sprintf("%s:%d", host, port)
```
smothiki: I'm not facing this issue. You might have to recheck why you want to do this here.

rootfs/etc/confd/templates/builder
```diff
@@ -109,6 +112,16 @@ sed -i -- "s#repo_name#$META_NAME#g" /etc/${SLUG_NAME}.yaml
...
+ACCESS_KEY=`cat /var/run/secrets/object/store/access-key-id`
+ACCESS_SECRET=`cat /var/run/secrets/object/store/access-secret-key`
+mkdir -p /var/minio-conf
+CONFIG_DIR=/var/minio-conf
+echo "/bin/mc -C $CONFIG_DIR config host add "http://$S3EP" $ACCESS_KEY $ACCESS_SECRET"
+/bin/mc -C $CONFIG_DIR --quiet config host add "http://$S3EP" $ACCESS_KEY $ACCESS_SECRET
+/bin/mc -C $CONFIG_DIR --quiet mb "${S3EP}/git"
+/bin/mc -C $CONFIG_DIR --quiet cp ${APP_NAME}.tar.gz $TAR_URL
+echo "stored tarfile in $TAR_URL"
+
```
smothiki: /bin is not required its in $PATH

pkg/etcd/etcd.go
arschles: ok, I'll check now

arschles: @smothiki can't reproduce it now, so I'll back out this change. I ran into the issue when I was setting everything up, so could have just been a n00b mistake.

PR #20, opened 2015-12-09
fix(deis-builder-rc.yaml): add imagePullPolicy: Always to rc manifest

arschles:


manifests/deis-builder-rc.yaml
```diff
@@ -14,7 +14,8 @@ spec:
         app: deis-builder
     spec:
       containers:
-        - name: deis-builder
+        - imagePullPolicy: Always
+          name: deis-builder
           image: quay.io/deisci/builder:v2-alpha
```
sgoings: any reason name: wouldn't go above the pull policy?

arschles: nope, I arbitrarily ordered the two

arschles: but now that you bring it up, `name` should go first. thoughts?

PR #22, opened 2015-12-10
fix(builder): use workflow v2 hook urls

Joshua-Anderson:
Ref deis/workflow#58

Are the workflow hooks used anywhere else?


arschles:
thanks @Joshua-Anderson !

PR #31, opened 2015-12-14
chore(Dockerfile): remove top level dockerfile

### LINKED ISSUE #19: do builds outside of the docker build
`go build boot.go` is currently run in the [top level Dockerfile](https://github.com/deis/builder/blob/master/Dockerfile), during a `docker build`. instead of doing this, it should just run as a `docker run` command. see https://github.com/deis/minio/pull/20 for an example

arschles:
It’s not used anywhere, as `go build`s are done in a simple `docker run` command from the Makefile. 

Fixes #19


arschles:
@technosophos @mboersma I'm gonna wait to merge this until all the `v2.0-alpha1

PR #34, opened 2015-12-15
fix(deis-builder-rc.yaml): fix indentation on deis-builder RC

arschles:
not marking showstopper because it's correct in the [helm chart](https://github.com/deis/charts/blob/master/deis/manifests/deis-builder-rc.yaml)


arschles:
@technosophos @mboersma I'm gonna wait to merge this until after the alpha showstoppers are in

PR #43, opened 2015-12-15
fix(Dockerfile,deis-builder-rc.yaml): add DOCKERIMAGE env var

arschles:
We have to set it so the minio client (https://github.com/minio/mc) works properly. Also, we can remove it in the RC manifest because it’s already in the Dockerfile.

The env var can also be removed from [deis/charts](https://github.com/deis/charts/blob/master/deis/manifests/deis-builder-rc.yaml#L30) once this is merged, but since it's redundant it's not a hard requirement for this PR.

### DIFF
--- a/manifests/deis-builder-rc.yaml
+++ b/manifests/deis-builder-rc.yaml
@@ -23,9 +23,6 @@ spec:
           env:
             - name: "EXTERNAL_PORT"
               value: "2223"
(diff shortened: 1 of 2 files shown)


technosophos:
This isn't a blocker for alpha, is it?

arschles:
nope, marked as beta

Issue #44, opened 2015-12-15
have the builders communicate to `builder` directly


arschles:
builder outputs the build status on a `git push`, so it would be nice if deis/dockerbuilder and deis/slugbuilder could communicate what they're doing back to builder, so the end-user (who did the `git push`) can see what's going on. currently they have no insight into what's going on while the build is in progress.

We'll likely need to finish part or all of #29 before this can be possible.

arschles:
this can probably be solved simply with running `kubectl logs -f` instead of having builder pods talk back to builder. it's currently done [in the current builder](https://github.com/deis/builder/blob/master/rootfs/etc/confd/templates/builder#L142)

PR #47, opened 2015-12-15
fix(builder): ping remote storage to determine when builds are done

### LINKED ISSUE #45: builder repeatedly prints out `no file` messages, and builds never finish, when using minio or other external object storage.
Even after slugbuilder finishes building and uploading the slug, the [builder script](https://github.com/deis/builder/blob/master/rootfs/etc/confd/templates/builder) will continue to check (on [line 139](https://github.com/deis/builder/blob/master/rootfs/etc/confd/templates/builder#L139)) for a local slugfile. While that works for fetcher configurations, it doesn't work for remote object storage configurations like Minio.

For Minio or remote object storage configurations, for `v2.0-alpha1` it should ping the object storage service to determine when the slug is uploaded. In a future version, https://github.com/deis/builder/issues/44 can remove the need for such pinging.

### LINKED ISSUE #49: Builder error on second push
Pushing an app a second time ran into at least two issues:
1. git bucketname already exists
2. pod name `deis-dockerbuilder` already existed

```console
example-dockerfile-http [master]$ g push deis master
Counting objects: 47, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (37/37), done.
Writing objects: 100% (47/47), 7.16 KiB | 0 bytes/s, done.
Total 47 (delta 6), reused 0 (delta 0)
Added host ‘http://10.3.0.133:9000’ successfully.
mc: <ERROR> Unable to make bucket ‘http://10.3.0.133:9000/git’. The requested bucket name is not available.
‘shadow-woodshed.tar.gz’ -> ‘http://10.3.0.133:9000/git/home/shadow-woodshed:git-3edec618/tar’
Total: 2.00 KB, Transferred: 2.00 KB, Speed: 801.59 KB/s
-----> stored tarfile in http://10.3.0.133:9000/git/home/shadow-woodshed:git-3edec618/tar
-----> creating builder pod in namespace deis
Error from server: error when creating "/etc/shadow-woodshed:git-3edec618.yaml": pods "deis-dockerbuilder" already exists
To ssh://git@a097f2098a2ce11e5a5a40272b026e4b-190952671.us-west-2.elb.amazonaws.com:2222/shadow-woodshed.git
 * [new branch]      master -> master
```

Also issues with deisci/dockerbuilder image
```
Failed to pull image "quay.io/deisci/dockerbuilder:v2-alpha": image pull failed for quay.io/deisci/dockerbuilder:v2-alpha, this may be because there are no credentials on this request.  details: (Error: Status 403 trying to pull repository deisci/dockerbuilder: "{\"error\": \"Permission Denied\"}")
```

arschles:
Only do so when builder is configured to upload to remote storage.

Fixes #45 
Fixes #49

### DIFF
--- a/rootfs/etc/confd/templates/builder
+++ b/rootfs/etc/confd/templates/builder
@@ -141,17 +141,21 @@ kubectl --namespace=$POD_NAMESPACE create -f /etc/${SLUG_NAME}.yaml
 
 if [[ "$REMOTE_STORAGE" == "1" ]]; then
   LS_CMD="$MC_PREFIX ls $PUSH_URL"
+  echo "Waiting for build to complete..."
   until $LS_CMD &> /dev/null; do
-    echo "no file"
+    echo -ne "."
     sleep 2
   done
 else
+  echo -ne "Waiting for build to complete..."
   while [ ! -f /apps/${SLUG_NAME}/slug.tgz ]
   do
+    echo -ne "."
     sleep 2
-    echo "no file"
   done
 fi
+echo ""
+echo "Build complete."
 
 URL="http://$DEIS_WORKFLOW_SERVICE_HOST:$DEIS_WORKFLOW_SERVICE_PORT/v2/hooks/config"
 RESPONSE=$(get-app-config -url="$URL" -key="{{ getv "/deis/controller/builderKey" }}" -user=$USER -app=$APP_NAME)
--- a/rootfs/etc/confd/templates/builder
+++ b/rootfs/etc/confd/templates/builder
@@ -98,13 +98,16 @@ fi
 git archive --format=tar.gz ${GIT_SHA} > ${APP_NAME}.tar.gz
 
 HTTP_PREFIX="http"
+REMOTE_STORAGE="0"
 # if minio is in the cluster, use it. otherwise use fetcher
 # TODO: figure out something for using S3 also
 if [[ -n "$DEIS_MINIO_SERVICE_HOST" && -n "$DEIS_MINIO_SERVICE_PORT" ]]; then
   S3EP=${DEIS_MINIO_SERVICE_HOST}:${DEIS_MINIO_SERVICE_PORT}
+  REMOTE_STORAGE="1"
 elif [[ -n "$DEIS_OUTSIDE_STORAGE_HOST" && -n "$DEIS_OUTSIDE_STORAGE_PORT" ]]; then
   HTTP_PREFIX="https"
   S3EP=${DEIS_OUTSIDE_STORAGE_HOST}:${DEIS_OUTSIDE_STORAGE_PORT}
+  REMOTE_STORAGE="1"
 elif [ -z "$S3EP" ]; then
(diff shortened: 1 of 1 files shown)


rootfs/etc/confd/templates/builder
```diff
@@ -136,11 +139,19 @@ kubectl --namespace=$POD_NAMESPACE create -f /etc/${SLUG_NAME}.yaml
...
+  until $LS_CMD &> /dev/null; do
+    echo "no file"
+    sleep 2
+  done
+else
+  while [ ! -f /apps/${SLUG_NAME}/slug.tgz ]
+  do
+    sleep 2
+    echo "no file"
+  done
```
smothiki: can we change this message to something else. Instead of no file a progress bar.

arschles: @smothiki we can remove it for now, but a progress bar doesn't make much sense because the build is binary from builder's perspective. I thing a better long term solution is #44

technosophos: This is a good enough message for alpha. It's useful for debugging. We can clean it up later if we have time. Please don't spend time on UI elements until the builder is working.

arschles: or are you referring to putting some dots to indicate to the user that things aren't frozen?

arschles:
after this fix is in, `git push`es seem to successfully deploy an app. below is the end result after slugbuilder finishes:

```console
[2015-12-15 23:54:55 UTC]     0B push/
-----> Launching... 
       done, jiggly-pendulum:v3 deployed to Deis

       http://jiggly-pendulum.localhost

       To learn more, use `deis help` or visit http://deis.io

To ssh://git@10.3.0.22:2222/jiggly-pendulum.git
 * [new branch]      master -> master
```

@smothiki - I added the progress indicator (I think this is what you were referring to) in https://github.com/deis/builder/commit/36607fa4d1c518374ae7c6b1bb0fcf70025f6a02. Also, do you know where the `[2015-12-15 23:54:55 UTC]     0B push/` comes from? It looks like output from a go program, but there are a few that get called after the build completes, so it would be helpful if you can narrow it down.

rootfs/etc/confd/templates/builder
arschles: @technosophos fyi I changed this in https://github.com/deis/builder/commit/36607fa4d1c518374ae7c6b1bb0fcf70025f6a02. It now continuously appends `.` characters to the end of the line, which is no less useful than what we have now (and takes up less screen real estate). My hope is that this will become better in the future.

technosophos: Saw that. That's fine. I just don't want these hugely important commits to go down rabbit holes with stuff like progress bars.

arschles: :+1:

smothiki:
I have no idea where push is coming from.

arschles:
@smothiki thanks. I've created https://github.com/deis/builder/issues/52 to track it

PR #61, opened 2015-12-18
fix(builder): add status messages

arschles:
Without these messages, the user may have to wait for a while before the builder pod spins up and the slug runner spins up (respectively).

This is an optional patch for `v2.0-alpha1`.

Fixes #64


slack:
This should goooo innnnnnnnnnnn, need another LGTM.

carmstrong:
LGTM3

slack:
3TGM, WHERE ARE YOU!>?!?!?

arschles:
not gonna merge this until after alpha is shipped ;)

slack:
What if I add the showstopper label?

arschles:
Oooh, you're playing hardball I see...

arschles:
it seems like so small a change that it couldn't hurt anything. I'm just paranoid after https://github.com/deis/minio/pull/39

slack:
What could go wrong!??!! I think the UX is bad enough that we should bring this in.

arschles:
@slack have you tested it recently? Last time I did was at least 10 days ago

arschles:
kidding..... pushing the green button

slack:
:champagne: !

PR #71, opened 2015-12-30
fix(Makefile,glide.lock): upgrade to glide 0.8 & add lockfile

arschles:
Fixes #70

PR #72, opened 2015-12-31
ref(*): consolidate builder into fewer Go binaries

### LINKED ISSUE #29: [Meta] consolidate into a single Go binary
This is a high level ticket to rewrite parts of builder in Go, and consolidate the various Go binaries into a single one.

We currently have many Go binaries that are built and used by shell scripts. This ticket is for:

1. Porting those shell scripts to Go
2. Moving the functionality in the external binary artifacts into the single builder process.

Since the shell scripts need the external binary artifacts, (1) must happen before (2).

When work on this starts, this ticket should be broken into smaller chunks.

### LINKED ISSUE #37: build a `boot` binary instead of a `builder` binary
the [prototype repo's `Makefile`](https://github.com/deis/prototype-repo/blob/master/Makefile#L52) does that, so this issue would bring builder in line with standards

### LINKED ISSUE #51: create builder pods with a uuid
# Problem
As seen in https://github.com/deis/builder/issues/49, if a build fails for some reason and the user's `git` client believes that the latest rev hasn't been pushed, it's reasonable to think that they will try to `git push` the same rev again.

In that case, builder will try to start a builder pod with the same name (because it computes the pod name based only on the rev). As a defensive measure, it would be useful for the builder pod to also include a random uuid so that even in this error case, the builder can re-build the same git rev.

# Workaround

Before this is fixed, you can make another commit, push again, and builder will successfully name the builder pod something new, based on the new git sha. For example, a `git commit --allow-empty -m "empty"` will do the trick.

# Issue Status
Not marking as showstopper, but keeping it in alpha. Treating this as a nice to have, but required for beta since we have a known workaround in the meantime.

arschles:
This is a big patch to refactor part of this component to Go from shell scripts. See [testing instructions below](https://github.com/deis/builder/pull/72#issuecomment-171004470).

# Guide to Changes

- Everything inside if `pkg/gitreceive` replaces the `rootfs/etc/confd/templates/builder` and `rootfs/etc/confd/templates/receive` scripts, which were invoked by the script generated by the `PreReceiveHookTpl` template in `pkg/git/git.go`. **This is by far the biggest change in this PR**
- The `Makefile` now only builds a `boot` binary
- `boot.go` now accepts the following commands:
  - `server` runs the SSH server (to accept `git push`es) and the fetcher, which is an old filesystem based object store
  - `git-receive` is run in the pre-receive hook by the SSH server in response to every `git push`
- Related to the previous, everything inside of `pkg/src` was previously used by the `rootfs/etc/confd/templates/builder` and `rootfs/etc/confd/templates/receiver` scripts, and was merged into code inside of `pkg/gitreceive`
- `pkg/git/git.go` was refactored to split the repo creation step and the pre-receive hook rendering step
- A log package was added, and is used inside of the git-receive hook (everything under `pkg/gitreceive`)
- Builder manifests were added to support public object storage systems

# TODOs

- [X] Run the SSH server and fetcher HTTP server in the same process (no more separate binary build for fetcher)
- [X] Also output stderr for the git-receive hook to auxiliary `io.Writer`. This functionality is currently used to collect stderr logs in a `bytes.Buffer` and then print out those logs on error
- [x] Rewrite the entire git-receive hook in Go. This work will consolidate all of the binaries that are built from the code in `pkg/src` into a single `gitreceive` binary
- [x] Add instructions in this PR on how to test (see https://github.com/deis/builder/pull/72#issuecomment-171004470)

As a result of the final bullet, the entire builder will be a single binary. The code changes involved to do that are below:

- [X] Refactor `boot.go` to accept 2 commands (uses [cli](https://github.com/codegangsta/cli) to build the commands) - `server` and `git-receive`. `server` starts the ssh server (and fetcher for now, but not after #6), and `git-receive` is started by the git post-receive hook
- [x] Add a `pkg/gitreceive` directory and package
- [x] Port the `rootfs/etc/confd/templates/{builder,receiver}` scripts to Go, and put them in `pkg/gitreceive`
- [X] Change the `Makefile` to just compile `boot.go` into a single `boot` executable, which still lives in `rootfs/bin/boot`
- [X] Change the `Dockerfile`'s entrypoint to include the `server` command on `boot`
- [X] Add a `log` package for pretty-printing logs
- [X] Add all necessary dependencies to the glide files
- [x] Change `preReceiveHookTplStr` to set the appropriate environment vars and call `/bin/boot git-receive`
- [x] Implement the etcd key fetching logic in the git receive hook (a function called `getBuilderKey`)
- [x] Port the changes in https://github.com/deis/builder/pull/78 to here
- [x] Port the applicable changes in https://github.com/deis/builder/pull/74 to here
- [x] Port the changes in https://github.com/deis/builder/pull/88 to here
- [ ] Remove the `remote:` prefixes. See [relevant gitreceive code](https://github.com/progrium/gitreceive/blob/0cc00c01aa6acb8a1db72d4b745675ed23ce591f/gitreceive#L123) for how it was done

Note that this PR will **not** be removing confd due to the `authorized_keys` file. See #83 (now replaced by #81) for that

Fixes #29 
Fixes #37 
Fixes #51
Depends on #71


arschles:
removing `awaiting review` because launching the app takes an unusually long time after this change. investigating

arschles:
logs from release start until the hang:

```console
10.168.0.1 "POST /v2/hooks/build HTTP/1.1" 200 56 "deis-builder"
10.168.0.1 "POST /v2/hooks/push HTTP/1.1" 201 423 "deis-builder"
10.168.0.1 "POST /v2/hooks/config HTTP/1.1" 200 194 "deis-builder"
INFO gotest: build gotest-7e00396 created
INFO gotest: release gotest-v4 created
INFO [gotest]: arschles deployed 816b0de
DEBUG deploy gotest_v4.web, img gotest:git-816b0de22a15a89e773a19ffe20830eb683d41f9, params {u'num': 0, u'tags': {}, u'envs': {}, u'cpu': {}, u'version': u'v4', u'memory': {}, u'aname': u'gotest'}, cmd "bash -c 'example-go'"
DEBUG scaling release gotest-v4-web to 1 out of final 1
```

arschles:
# Testing Instructions

Assuming you have no deis cluster installed:

- `helm repo add https://github.com/deis/charts`
- `helm fetch deis/deis-dev`
- Edit the ~/.helm/workflow/charts/deis-dev/manifests/deis-builder-rc.yaml to use the quay.io/arschles/builder:go-rewrite image
  - You can also add `DEBUG=true` to the env to see debug output on the builder
- `helm install deis-dev`
- `git clone https://github.com/deis/example-go.git && cd example-go`
- `deis register deis.$INGRESS_IP.xip.io` (use the email address in your `~/.ssh/id_rsa.pub`)
- `deis keys:add ~/.ssh/id_rsa.pub`
- `git remote remove deis`
- `deis create gotest`
- `git push deis master`
- `curl gotest.$INGRESS_IP.xip.io` to see the newly deployed app in action

krancour:
Whatever you do, do not squash these commits anytime soon.  This is a lot easier to review commit-by-commit.

arschles:
@krancour I certainly won't.

Speaking of which, I'll add some text breaking this change down into subsections. Hopefully that'll make it easier to review.

pkg/gitreceive/controller.go
```diff
@@ -0,0 +1,150 @@
...
+	req, err := http.NewRequest("POST", url, b)
+	if err != nil {
+		return nil, err
+	}
+
+	setReqHeaders(builderKey, req)
+
+	log.Debug("Workflow request POST /v2/hooks/config\n%s", string(data))
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
```
smothiki: Is this working?  Last time we discussed , you faced an issue here and decided to use get request.

smothiki:
Looks good will add LGTM once I test manually and I want to revisit code again

pkg/gitreceive/controller.go
arschles: yep, as long as we send `"image"` as the same value as the app name, it works. the `GET` request has a totally different authentication code path in the controller, according to @helgi, so it's not a good option for builder

helgi:
https://github.com/deis/builder/pull/93 needs to be applied to this PR

arschles:
@helgi can do. is the latest v2-beta workflow image compatible with that change? I'd like to ensure this PR is testable with the stock deis-dev chart

arschles:
@helgi nvm, I see that deis/charts#92 is in. I have changes to port #93 over, and will test and push shortly.

pkg/gitreceive/build.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+	}
+
+	log.Info("Build complete.")
+
+	log.Info("Launching...")
+
+	buildHook := &pkg.BuildHook{
+		Sha:         gitSha,
+		ReceiveUser: conf.Username,
```
smothiki: can you just send short sha here

arschles: I'm assuming you mean the line above. Unfortunately I can't - workflow sends back a `400 Bad Request` when I do...

pkg/gitreceive/build.go
```diff
@@ -0,0 +1,295 @@
...
+	if err := os.MkdirAll(buildDir, os.ModeDir); err != nil {
+		return fmt.Errorf("making the build directory %s (%s)", buildDir, err)
+	}
+	tmpDir := os.TempDir()
+
+	tarURL := fmt.Sprintf("%s://%s:%s/git/home/%s/tar", storage.schema(), storage.host(), storage.port(), slugName)
+
+	// this is where workflow tells slugrunner to download the slug from, so we have to tell slugbuilder to upload it to here
+	pushURL := fmt.Sprintf("%s://%s:%s/git/home/%s/push", storage.schema(), storage.host(), storage.port(), fmt.Sprintf("%s:git-%s", appName, gitSha))
+
```
smothiki: git sha to short sha here as well

pkg/gitreceive/build.go
bacongobbler: The controller will shorten the sha. It's best to just use the entire git sha here, which then the controller will shorten in API views for presentation purposes.

smothiki: https://github.com/deis/builder/blob/master/rootfs/etc/confd/templates/builder#L189. controller accepts short sha . Is there any recent change in workflow that is not accepting short sha's

pkg/gitreceive/build.go
arschles: Unfortunately that causes slugrunner to fail to download the slug. See comment one line above this line.

pkg/gitreceive/build.go
smothiki: Ok

arschles: @smothiki I'm not sure RE changes in controller, but I've tried in this code sending both the short and long sha, and it appears that only the long sha produces a successful result from workflow.

pkg/gitreceive/build.go
smothiki: I'll change the URL format in a different PR . The URL we are uploading is meant for fetcher. I'll create an issue for this

arschles: ok, thanks!

arschles: @smothiki actually, I think it would be nicer if the controller API `/v2/hooks/build` could take the URL to assign to slugrunner. It would reduce the coupling between builder and the controller.

pkg/git/git.go
```diff
@@ -160,9 +149,7 @@ func plumbCommand(cmd *exec.Cmd, channel ssh.Channel, sidechannel io.Writer) *sy
 	}()
 
 	cmd.Stdout = channel
-	cmd.Stderr = channel.Stderr()
-
-	return &wg
```
technosophos: I take it the old waitgroup add calls were all removed at some intermediary point? At one point, the purpose of the wait group was to make sure that for large inputs, we could read in the background, but parallelize calls at a higher level. It looks like this may have eroded away over time, but I'm concerned that a big input could have some weird issues if this copy is still going on at the same time that sidechannel gets consumed.

More specifically, if this is 3/4 of the way through a write on the io.Copy operation, and you read to the end of the sidechannel file, will you get an EOF, or will it start blocking?

arschles: >I take it the old waitgroup add calls were all removed at some intermediary point

It appears so, because `plumbCommand` didn't modify this `WaitGroup`, nor did any caller. The one caller of this `func` called `Wait` on the `WaitGroup`, which was effectively a no-op because there was never any `Add` call

>At one point, the purpose of the wait group was to make sure that for large inputs, we could read in the background, but parallelize calls at a higher level. It looks like this may have eroded away over time, but I'm concerned that a big input could have some weird issues if this copy is still going on at the same time that sidechannel gets consumed.

Makes sense, but where would we need to parallelize at a higher level? This and the calling `func Receive` is already running in a goroutine. Anyway, the `cmd.Wait` in the aforementioned `func Receive` will block progress before the sidechannel is consumed (or, in practical terms, `errbuf.Bytes()` is called). See anything I'm missing?

>More specifically, if this is 3/4 of the way through a write on the io.Copy operation, and you read to the end of the sidechannel file, will you get an EOF, or will it start blocking?

Excellent question. The `io.Copy` just copies all input from the `ssh.Channel` to the command's STDIN, so it won't have any effect on the sidechannel `io.Writer`. As soon as the `ssh.Channel` returns `EOF` from its `Read()`, the copy will return with no error and then stdin pipe will close. From testing, that doesn't seem to be an issue for the `git-shell` command. Could I be wrong?

However, there is a bug here because as soon as data starts coming in on STDIN to the command, it may start writing to STDOUT or STDERR, but we haven't assigned `io.Writer`s to `cmd.Stdout` or `cmd.Stderr`. I'll push a fix for that.

technosophos: Good catch on that STDOUT/STDERR issue.

Okay, now I remember how this code works. Your explanation makes sense to me. I think this is good.

arschles: You got me thinking. I think there's a simpler way to do this without the extra goroutine - start the command, then `io.Copy`, then wait. Something like this:

```go
	inpipe, err := cmd.StdinPipe()
	if err != nil {
		return nil, err
	}
	cmd.Stdout = channel
	cmd.Stderr = io.MultiWriter(channel.Stderr(), &errbuff)

	if err := cmd.Start(); err != nil {
		err = fmt.Errorf("Failed to start git pre-receive hook: %s (%s)", err, errbuff.Bytes())
		log.Warnf(c, err.Error())
		return nil, err
	}

	if _, err := io.Copy(inpipe, channel); err != nil {
		err = fmt.Errorf("Failed to write git objects into the git pre-receive hook (%s)", err)
		log.Warnf(c, err.Error())
		return nil, err
	}

	fmt.Println("Waiting for git-receive to run.")
	fmt.Println("Waiting for deploy.")
	if err := cmd.Wait(); err != nil {
		err = fmt.Errorf("Failed to run git pre-receive hook: %s (%s)", errbuff.Bytes(), err)
		log.Errf(c, err.Error())
		return nil, err
	}
```

I'm testing the above code now with a 400+mb file.

pkg/log/log.go
```diff
@@ -0,0 +1,63 @@
+// Package log is a convenience wrapper for logging messages of various levels (associated colors to come)
+// to the terminal. Much of this code has been shamelessly stolen from https://github.com/helm/helm/blob/master/log/log.go
```
technosophos: lol... we need to write a decent implementation of this and put it in `deis/pkg`. I just nabbed the same piece of helm to mock up something else.

arschles: @technosophos in progress at https://github.com/deis/pkg/pull/9

pkg/gitreceive/build.go
bacongobbler: +1 for reduced coupling

pkg/gitreceive/build.go
```diff
@@ -0,0 +1,295 @@
...
+	// Ensure that the app config can be gotten from workflow. We don't do anything with this information
```
bacongobbler: Aren't we doing something if we're searching for a custom buildpack URL? ;)

pkg/gitreceive/build.go
```diff
@@ -0,0 +1,295 @@
...
+	appConf, err := getAppConfig(conf, builderKey, conf.Username, appName)
+	if err != nil {
+		return fmt.Errorf("getting app config for %s (%s)", appName, err)
+	}
+	log.Debug("got the following config back for app %s: %+v", appName, *appConf)
+	var buildPackURL string
+	if buildPackURLInterface, ok := appConf.Values["BUILDPACK_URL"]; ok {
+		if bpStr, ok := buildPackURLInterface.(string); ok {
```
bacongobbler: wouldn't the variable `ok` already be in use in this context? Maybe my Go fu isn't that good after all.

pkg/gitreceive/build.go
```diff
@@ -0,0 +1,295 @@
...
+			log.Debug("found custom buildpack URL %s", bpStr)
+			buildPackURL = bpStr
+		}
+	}
+
+	// build a tarball from the new objects
+	gitArchiveCmd := repoCmd(repoDir, "git", "archive", "--format=tar.gz", fmt.Sprintf("--output=%s.tar.gz", appName), gitSha)
```
bacongobbler: appname.tar.gz is being used in a couple places (line 117). Should we make a variable just for best practice?

pkg/gitreceive/build.go
arschles: you're right, it's already there but is shadowed inside the inner `if` statement only

pkg/gitreceive/build.go
arschles: oops, outdated comment - I'll remove. good catch!

pkg/gitreceive/build.go
```diff
@@ -0,0 +1,295 @@
...
+	"time"
+
+	"code.google.com/p/go-uuid/uuid"
+	"github.com/deis/builder/pkg"
+	"github.com/deis/builder/pkg/log"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	shortShaIdx = 8
```
bacongobbler: mind adding a comment here to explain what this number represents?

pkg/gitreceive/build.go
arschles: sure, I'll do so

pkg/gitreceive/build.go
bacongobbler: neat

pkg/gitreceive/build.go
arschles: sure, will do

pkg/gitreceive/build.go
```diff
@@ -0,0 +1,295 @@
...
+	tarCmd.Stdout = os.Stdout
+	tarCmd.Stderr = os.Stderr
+	if err := run(tarCmd); err != nil {
+		return fmt.Errorf("running %s (%s)", strings.Join(tarCmd.Args, " "), err)
+	}
+
+	usingDockerfile := true
+	rawProcFile, err := ioutil.ReadFile(fmt.Sprintf("%s/Procfile", tmpDir))
+	if err == nil {
+		usingDockerfile = false
```
bacongobbler: what about apps that have both a Dockerfile and a Procfile? In v1 we assumed that this app is still using a Dockerfile, but the Procfile would be used as the source of truth for process types. It's weird, I know...

arschles: I wasn't really sure, but matched what the old builder was doing. I'd like to leave as-is in this PR, but if I should be doing something different, can you file an issue?

bacongobbler: Absolutely!

rootfs/etc/confd/templates/builder
```diff
@@ -1,208 +0,0 @@
...
-TMP_DIR=$(mktemp -d -p $BUILD_DIR)
-
-cd $REPO_DIR
-# use Procfile if provided, otherwise try default process types from ./release
-git archive --format=tar.gz ${GIT_SHA} > ${APP_NAME}.tar.gz
-tar -xzf ${APP_NAME}.tar.gz -C $TMP_DIR/
-USING_DOCKERFILE=true
-if [ -f $TMP_DIR/Procfile ]; then
-    PROCFILE=$(cat $TMP_DIR/Procfile | yaml2json-procfile)
-    USING_DOCKERFILE=false
```
bacongobbler: huh, that's certainly interesting. Yes I'll file a bug about this, @arschles

arschles: thanks @bacongobbler

technosophos:
Okay, I've read the code from top to bottom, and I'm LGTMing this. Whoever gives LGTM2 should probably also do some manual testing, since this thing is huge.

bacongobbler:
> Whoever gives LGTM2 should probably also do some manual testing, since this thing is huge.

on it :+1:

arschles:
Thanks @technosophos & @bacongobbler

pkg/gitreceive/build.go
```diff
@@ -0,0 +1,295 @@
...
+	if err != nil {
+		return fmt.Errorf("publishing release (%s)", err)
+	}
+	release, ok := buildHookResp.Release["version"]
+	if !ok {
+		return fmt.Errorf("No release returned from Deis controller")
+	}
+
+	log.Info("Done, %s:v%d deployed to Deis", appName, release)
+	log.Info("Use 'deis open' to view this application in your browser")
```
bacongobbler: I believe the old layout was

```
done, ...

use deis open...

to learn more...
```

I'm going to test manually, but wouldn't the output now be

```
done, ...
use deis open...
to learn more...
```

Since there's a lack of extra newline characters? I know our BDFL was quite stingy on the format

arschles: Yep, good catch. I'll add newlines back in right now.

bacongobbler:
Is quay.io/arschles/builder:go-rewrite up to date with all the new changes? happy to build my own image if this isn't automated :)

arschles:
@bacongobbler that image is not up to date, but it's pushing now. If you want to build yourself, here's the command: 

```console
IMAGE_PREFIX=$MY_REPO_NAME VERSION=go-rewrite make build docker-build docker-push
```

arschles:
@bacongobbler ok, `quay.io/arschles/builder:go-devel` is up to date now. 

Also, to answer your question, building images for this PR isn't automated.

pkg/gitreceive/build.go
```diff
@@ -0,0 +1,297 @@
...
+	// https://github.com/deis/builder/issues/80 will eliminate this distaste
+	run(makeBucketCmd)
+
+	cpCmd := mcCmd(configDir, "cp", appTgz, tarURL)
+	cpCmd.Dir = repoDir
+	if err := run(cpCmd); err != nil {
+		return fmt.Errorf("copying %s to %s (%s)", appTgz, tarURL, err)
+	}
+
+	log.Info("Starting build")
```
bacongobbler: There's a long pause right here while we're waiting to boot the pod. Why not something like `Starting build... but first, coffee!` so people know that it's intentionally going to take a while

bacongobbler:
My build was caught in a loop forever with the following:

```
remote: mc: <ERROR> Unable to stat ‘http://10.247.115.24:9000/git/home/go:git-5450cbcdaaf9afe6fadd219c94ac9c449bd62413/push’. The specified key does not exist.
```

My guess is that it we should use the short sha here?

Also seems like the ol' `remote: ` prefix is back. History on this bug here: https://github.com/deis/deis/issues/1216

pkg/gitreceive/build.go
arschles: could do that, but I'd rather do it in a future PR, so that the logs don't change from what was previously here

bacongobbler: okay I'll file another issue

arschles:
@bacongobbler I'll look at removing the remote prefix. Also, are you up to date on eveything, and running on, the `deis-dev` chart?

That endless loop is likely not related to the short sha issue - instead, it's more likely that the slugbuilder didn't upload the slug to the expected location because it failed. Do you see any logs above that indicate the failure?

bacongobbler:
Looks like I am using 27fd67fa0c0b06766de12120c72eb433039f2744 so I'm out of date. I'll rebase and try again later.

> Do you see any logs above that indicate the failure?

I dunno. My screen just started dumping that message like no tomorrow, which was enough to hit my scrollback limit. I'll try another build and see if I get anything useful. It's likely that me being outdated could have caused that issue.

arschles:
thanks @bacongobbler

bacongobbler:
@arschles for reference, this is what stripped the remote prefix in the old gitreceive: https://github.com/progrium/gitreceive/blob/0cc00c01aa6acb8a1db72d4b745675ed23ce591f/gitreceive#L119-L124

arschles:
@bacongobbler thanks, that saves me a ton of searching

pkg/gitreceive/build.go
bacongobbler: #95

pkg/gitreceive/build.go
bacongobbler: #94

arschles: thanks!

arschles:
@bacongobbler ok, I added the strip prefix logic back in https://github.com/arschles/builder/commit/3ac4e4834aa7139f4404e57f9a17d774c1d6cebe. Sample output from a `git push` is below. Notice that warnings & errors still have the prefix, which I believe is ok. Let me know if you disagree.

```console
ENG000656:example-go aaronschlesinger$ git commit -m "asd" --allow-empty && git push deis master
[master 27a04f1] asd
Counting objects: 85, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (80/80), done.
Writing objects: 100% (85/85), 18.28 KiB | 0 bytes/s, done.
Total 85 (delta 40), reused 0 (delta 0)
remote: mc: <ERROR> Unable to make bucket ‘http://10.171.250.102:9000/git’. The requested bucket name is not available.
-----> Starting build
-----> Go app detected
-----> Checking Godeps/Godeps.json file.
-----> Installing go1.4.2... done
-----> Running: godep go install -tags heroku ./...
-----> Discovering process types
       Procfile declares types -> web
-----> Compiled slug size is 1.7M
-----> Build complete.
-----> Launching app.
-----> Launching...
-----> Done, gotest:v1 deployed to Deis

-----> Use 'deis open' to view this application in your browser

-----> To learn more, use 'deis help' or visit http://deis.io

To ssh://git@deis.130.211.154.75.xip.io:2222/gotest.git
 * [new branch]      master -> master
```

bacongobbler:
I'm on the latest version (`192.168.0.18:5000/deis/builder:git-74ded2e`) and I'm still seeing minio errors:

```><> git push deis master
Counting objects: 75, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (40/40), done.
Writing objects: 100% (75/75), 18.20 KiB | 0 bytes/s, done.
Total 75 (delta 30), reused 75 (delta 30)
remote: mc: <ERROR> Unable to make bucket ‘http://10.247.110.165:9000/git’. The requested bucket name is not available.
-----> Starting build
-----> Go app detected
remote: mc: <ERROR> Unable to stat ‘http://10.247.110.165:9000/git/home/foo:git-5450cbcdaaf9afe6fadd219c94ac9c449bd62413/push’. The specified key does not exist.
```

Logs from minio:

```
><> kd logs deis-minio-al13u
minio server /home/minio/

AccessKey: 8TZRY2JRWMPT6UMXR6I5  SecretKey: gbstrOvotMMcg2sMfGUhA5a6Et/EI5ALtIHsobYk

To configure Minio Client.

        $ wget https://dl.minio.io/client/mc/release/linux-amd64/mc
        $ chmod 755 mc
        $ ./mc config host add http://localhost:9000 8TZRY2JRWMPT6UMXR6I5 gbstrOvotMMcg2sMfGUhA5a6Et/EI5ALtIHsobYk
        $ ./mc mb localhost/photobucket
        $ ./mc cp ~/Photos... localhost/photobucket

Starting minio server:
Listening on http://127.0.0.1:9000
Listening on http://10.246.28.79:9000
time="2016-01-14T19:03:55Z" level=error msg="MakeBucket failed." Error={Bucket exists: git fs.BucketExists [{125 pkg/fs/fs-bucket.go fs.Filesystem.MakeBucket map[]} {255 bucket-handlers.go main.CloudStorageAPI.PutBucketHandler map[]}] map[host.lang:go1.5.2 mem.total:9.3MB mem.heap.used:1.9MB host.arch:amd64 host.name:deis-minio-al13u mem.used:1.9MB mem.heap.total:5.7MB host.os:linux host.cpus:4]} 
time="2016-01-14T19:04:37Z" level=error msg="MakeBucket failed." Error={Bucket exists: git fs.BucketExists [{125 pkg/fs/fs-bucket.go fs.Filesystem.MakeBucket map[]} {255 bucket-handlers.go main.CloudStorageAPI.PutBucketHandler map[]}] map[host.os:linux host.cpus:4 mem.total:9.3MB mem.heap.used:1.9MB mem.heap.total:5.7MB host.arch:amd64 host.name:deis-minio-al13u host.lang:go1.5.2 mem.used:1.9MB]}
```

arschles:
@bacongobbler the first error from the `git push` logs is expected (builder tries to create the bucket it needs every time it runs), but the second of course does not. Do logs continue after the second error? If so, I'm assuming they loop forever on the `mc ls` command?

Finally, can you run the builder with `DEBUG="true"` and run again?

bacongobbler:
Yep, here's a gist: https://gist.github.com/bacongobbler/3ea3cd8362e542d649aa

It loops forever on the `mc ls` command, yeah

arschles:
Ok, let's see what the output is when `DEBUG="true"` is on. I fixed an issue a few days ago that was causing the endless `mc ls` loop, and I haven't been able to repro since.

bacongobbler:
Unrelated networking issues with my cluster. LGTM! :shipit:

PR #74, opened 2016-01-04
feat(builder): adjust dockerbuilder and slugrunner templates

smothiki:
Modify dockerbuilder and slugbuilder templates if the secret is not present.

### DIFF
--- a/rootfs/etc/confd/templates/builder
+++ b/rootfs/etc/confd/templates/builder
@@ -84,10 +84,10 @@ fi
 
 if [[ ! -f /var/run/secrets/object/store/access-key-id ]]; then
   if $USING_DOCKERFILE ; then
-    l1=`grep -n "object-store" /etc/deis-dockerbuilder.yaml | head -n1 |cut -d ":" -f1`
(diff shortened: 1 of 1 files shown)


arschles:
@smothiki do we want to support running object storage without auth? IMO the answer is no, do you have a use case?

smothiki:
For fetcher which runs with out auth for now. Also if we want to give customizable storage solutions or outside storage which doesn't have auth then this will be helpful

arschles:
> For fetcher which runs with out auth for now

We agreed that Fetcher is gonna go away (#6)

>Also if we want to give customizable storage solutions or outside storage which doesn't have auth then this will be helpful

IMO we shouldn't encourage anyone to run object storage (especially not outside storage, but also in-cluster storage) without auth

smothiki:
In this PR I'm just fixing secrets as we have changed from object-store to minio-user. If we are thinking enforcing auth strictly I'll create different PR

arschles:
Ok, makes sense. I assumed wrong about what this PR was going to be. Sorry for jumping the gun, and let me know when you're ready for a review.

smothiki:
I have tested it and commands are working this PR ready for a review

PR #78, opened 2016-01-05
feat(builder): add support for custom buildpack url

smothiki:
Added buildpack url environment variable to slugbuilder template

### DIFF
--- a/pkg/src/get-app-config.go
+++ b/pkg/src/get-app-config.go
@@ -29,10 +29,11 @@ func main() {
 	builderKey := flag.String("key", "", "Builder Key")
 	user := flag.String("user", "", "Controller username")
 	app := flag.String("app", "", "Controller application name")
+	value := flag.String("value", "", "key value from config values")
 
 	flag.Parse()
 
-	if flag.NFlag() < 4 {
+	if flag.NFlag() < 5 {
 		flag.Usage()
 		os.Exit(1)
 	}
@@ -104,6 +105,18 @@ func main() {
 		fmt.Printf("%v\n", err)
 		os.Exit(1)
 	}
+
 	toString, err := json.Marshal(config)
-	fmt.Println(string(toString))
+	if err != nil {
+		fmt.Printf("failed JSON marshal (%v)\n", err)
+		os.Exit(1)
+	}
+
+	if *value == "all" {
+		fmt.Println(string(toString))
+	}
+	val, present := config.Values[*value]
+	if present {
+		fmt.Println(val)
+	}
 }
--- a/rootfs/etc/confd/templates/builder
+++ b/rootfs/etc/confd/templates/builder
@@ -74,6 +74,14 @@ else
     PROCFILE="{}"
 fi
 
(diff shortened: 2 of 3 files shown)


pkg/src/get-app-config.go
```diff
@@ -105,5 +106,12 @@ func main() {
 		os.Exit(1)
 	}
 	toString, err := json.Marshal(config)
```
arschles: Can you please add an error check for this returned error? I know this isn't code that you changed, but it would be nice to add it here.

arschles:
@smothiki I did a quick review just now, let me know when you're ready for a final review.

smothiki:
@arschles  done

pkg/src/get-app-config.go
```diff
@@ -104,6 +105,20 @@ func main() {
 		fmt.Printf("%v\n", err)
 		os.Exit(1)
 	}
+
 	toString, err := json.Marshal(config)
-	fmt.Println(string(toString))
+	if err != nil {
+		fmt.Println("failed JSON marshal")
+		fmt.Printf("%v\n", err)
```
arschles: can you merge these two lines. for example:

```go
fmt.Printf("failed JSON marshal (%v)\n", err)
```

pkg/src/get-app-config.go
```diff
@@ -104,6 +105,20 @@ func main() {
...
+		os.Exit(1)
+	}
+
+	if *value == "all" {
+		fmt.Println(string(toString))
+	} else {
+		_, present := config.Values[*value]
+		if present {
+			fmt.Println(config.Values[*value])
+		}
```
arschles: can simplify this:

```go
val, present := config.Values[*value]
if present {
  fmt.Println(val)
}
```

pkg/src/get-app-config.go
```diff
@@ -104,6 +105,20 @@ func main() {
...
+		fmt.Printf("%v\n", err)
+		os.Exit(1)
+	}
+
+	if *value == "all" {
+		fmt.Println(string(toString))
+	} else {
+		_, present := config.Values[*value]
+		if present {
+			fmt.Println(config.Values[*value])
```
arschles: this looks indented twice, can you remove one indentation?

rootfs/etc/confd/templates/builder
```diff
@@ -74,6 +74,14 @@ else
     PROCFILE="{}"
 fi
 
+URL="http://$DEIS_WORKFLOW_SERVICE_HOST:$DEIS_WORKFLOW_SERVICE_PORT/v2/hooks/config"
```
arschles: @smothiki does workflow support this endpoint?

arschles:
@smothiki one more of my comments left, but it doesn't block this PR. Can you remove `in progress` and add `awaiting review`, and let me know when you do? I'll add my LGTM then.

rootfs/etc/confd/templates/builder
smothiki: Yes. https://github.com/deis/workflow/blob/master/client/controller/models/config/config.go#L13
https://github.com/deis/workflow/blob/master/rootfs/api/urls.py#L79

arschles: thanks

PR #98, opened 2016-01-14
fix(Dockerfile): merging 2 apk commands

arschles:
docker builds are slightly faster this way


bacongobbler:
+1 saw this too but forgot to file an issue.

PR #103, opened 2016-01-15
chore(deploy.sh): Stop producing v2-beta images for each PR merge

jchauncey:
Instead we should produce a canary image. This allows us to cut stable release images while providing an image that tracks HEAD. This image should also be in the deis org.


arschles:
I saw discussion about this change elsewhere, are we going ahead with this change?

FWIW, I'm :+1: on it

PR #115, opened 2016-01-21
ref(*): use the kubernetes client directly (WIP)

### LINKED ISSUE #65: Fix race condition while attempting to stream logs from the builder pod.
https://github.com/deis/builder/blob/daac4b822a94554799c8503e5e7662592492b7f0/rootfs/etc/confd/templates/builder#L139

Is very prone to race conditions. If the builder pod either never makes it to `Running` or transitions through `Running` to `Terminated` we might hang forever.

### LINKED ISSUE #79: use Go templates in manifests
after #72 is done, render templates instead of doing string replaces

arschles:
Closes #79
Fixes #65 

Simplifies dependencies and the code

### DIFF
--- a/pkg/gitreceive/build.go
+++ b/pkg/gitreceive/build.go
@@ -1,14 +1,12 @@
 package gitreceive
 
 import (
-	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
-	"time"
 
 	"github.com/deis/builder/pkg"
 	"github.com/deis/builder/pkg/gitreceive/log"
@@ -16,6 +14,7 @@ import (
 	"gopkg.in/yaml.v2"
 	"k8s.io/kubernetes/pkg/api"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
+	"k8s.io/kubernetes/pkg/fields"
 )
 
 const (
@@ -158,7 +157,7 @@ func build(conf *Config, kubeClient *client.Client, builderKey, gitSha string) e
 	if usingDockerfile {
 		pod = dockerBuilderPod(
 			conf.Debug,
-			storageCreds != nil,
+			creds != nil,
 			dockerBuilderPodName(appName, shortSha),
 			conf.PodNamespace,
 			"deis",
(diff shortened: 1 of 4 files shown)


pkg/gitreceive/build.go
```diff
@@ -201,53 +152,72 @@ func build(conf *Config, builderKey, gitSha string) error {
...
-	// TODO: use the k8s client and watch the event stream instead (https://github.com/deis/builder/issues/65)
-	for {
-		cmd := kGetCmd(conf.PodNamespace, buildPodName)
-		var out bytes.Buffer
-		cmd.Stdout = &out
-		// ignore errors
-		run(cmd)
-		outStr := string(out.Bytes())
-		if strings.Contains(outStr, "phase: Running") {
+	watcher, err := podsInterface.Watch(labels.Nothing(), fields.OneTermEqualSelector("name", pod.Name), v1Version)
```
aledbf: @arschles which version of k8s is being targeted in V2?
I ask because this interface is different in v1.2

arschles: @aledbf In builder, and generally, we're targeting Kubernetes v1.1 for Deis v2 beta.

Sidenote: Some components like [deis/logger](https://github.com/deis/logger) will require [the DaemonSets extension](https://github.com/kubernetes/kubernetes/blob/master/docs/design/daemon.md).

pkg/gitreceive/k8s_util.go
```diff
@@ -0,0 +1,54 @@
...
+	return &api.Pod{
+		TypeMeta: unversioned.TypeMeta{Kind: podKind, APIVersion: v1Version},
+		ObjectMeta: api.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels: map[string]string{
+				"heritage": heritageLabel,
+				"version":  versionLabel,
+			},
+		},
```
smothiki: Where are we setting environment variables ?

arschles: We aren't yet but it'll have to be done in the below `Spec`. Same in `slugbuilderPod` below

smothiki: this is in work in progress  ?

arschles: yes, very much so

aledbf:
Closed by #133

PR #131, opened 2016-01-27
ref(build.go): pulling slugbuilder storage params out

### LINKED ISSUE #130: Builder is uploading slugs to wrong URL in object storage
ATM, builder is uploading built slugs to `http://<object storage>/git/home/<app>:git-<short sha>/tar`

But workflow tells the slugrunner to get the slug from `http://<object storage>/git/home/<app>:git-<long sha>/push/slug.tgz slug.tgz`

For example, a slug gets pushed to here:

`http://10.3.0.228:9000/git/home/scenic-waterbed:git-afb7db0a/tar`

And pulled from here:

`http://10.3.0.228:9000/git/home/scenic-waterbed:git-afb7db0a08510b74abbfc52a9e9745971190a37c/push/slug.tgz`

cc @arschles 

__EDIT:__ Apparently the former URL isn't actually for the slug, but for tarred source.  If that's the case, then the slug is actually completely missing from object storage after the build.

arschles:
and representing the git sha in a struct to do so

Fixes #130 
Makes progress on #132

### DIFF
--- a/pkg/gitreceive/git/sha.go
+++ b/pkg/gitreceive/git/sha.go
@@ -22,13 +22,16 @@ func (e ErrInvalidGitSha) Error() string {
 }
 
 type SHA struct {
-	Full  string
-	Short string
+	full  string
+	short string
 }
 
 func NewSha(rawSha string) (*SHA, error) {
 	if !shaRegex.Match([]byte(rawSha)) {
 		return nil, ErrInvalidGitSha{sha: rawSha}
 	}
-	return &SHA{Full: rawSha, Short: rawSha[0:8]}, nil
+	return &SHA{full: rawSha, short: rawSha[0:8]}, nil
 }
+
+func (s SHA) Full() string  { return s.full }
+func (s SHA) Short() string { return s.short }
--- a/pkg/gitreceive/git/sha.go
+++ b/pkg/gitreceive/git/sha.go
@@ -1,0 +1,31 @@
+package git
+
+import (
+	"fmt"
+)
+
+const (
+	// this constant represents the length of a shortened git sha - 8 characters long
+	shortShaIdx = 8
+	fullShaLen  = 40
+)
+
+type ErrGitShaTooShort struct {
+	sha string
+}
+
+func (e ErrGitShaTooShort) Error() string {
+	return fmt.Sprintf("git sha %s was too short", e.sha)
+}
+
+type SHA struct {
+	Full  string
+	Short string
+}
+
+func NewSha(rawSha string) (*SHA, error) {
+	if len(rawSha) < fullShaLen {
+		return nil, ErrGitShaTooShort{sha: rawSha}
+	}
+	return &SHA{Full: rawSha, Short: rawSha[0:8]}, nil
+}
--- a/pkg/gitreceive/git/sha.go
+++ b/pkg/gitreceive/git/sha.go
@@ -8,7 +8,6 @@ import (
 const (
 	// this constant represents the length of a shortened git sha - 8 characters long
 	shortShaIdx = 8
-	fullShaLen  = 40
 )
 
 var shaRegex = regexp.MustCompile(`^[\da-f]{40}$`)
--- a/pkg/gitreceive/git/sha.go
+++ b/pkg/gitreceive/git/sha.go
@@ -2,6 +2,7 @@ package git
 
 import (
 	"fmt"
+	"regexp"
 )
 
 const (
@@ -10,12 +11,14 @@ const (
 	fullShaLen  = 40
 )
 
-type ErrGitShaTooShort struct {
+var shaRegex = regexp.MustCompile(`^[\da-f]{40}$`)
+
+type ErrInvalidGitSha struct {
 	sha string
 }
 
-func (e ErrGitShaTooShort) Error() string {
-	return fmt.Sprintf("git sha %s was too short", e.sha)
+func (e ErrInvalidGitSha) Error() string {
+	return fmt.Sprintf("git sha %s was invalid", e.sha)
 }
 
 type SHA struct {
@@ -24,8 +27,8 @@ type SHA struct {
 }
 
 func NewSha(rawSha string) (*SHA, error) {
-	if len(rawSha) < fullShaLen {
-		return nil, ErrGitShaTooShort{sha: rawSha}
+	if !shaRegex.Match([]byte(rawSha)) {
+		return nil, ErrInvalidGitSha{sha: rawSha}
 	}
 	return &SHA{Full: rawSha, Short: rawSha[0:8]}, nil
 }
--- a/pkg/gitreceive/git/sha.go
+++ b/pkg/gitreceive/git/sha.go
@@ -27,7 +27,7 @@ type SHA struct {
 }
 
 func NewSha(rawSha string) (*SHA, error) {
-	if !shaRegex.Match([]byte(rawSha)) {
+	if !shaRegex.MatchString(rawSha) {
 		return nil, ErrInvalidGitSha{sha: rawSha}
 	}
 	return &SHA{full: rawSha, short: rawSha[0:8]}, nil
--- a/pkg/gitreceive/storage/slug_builder_info.go
+++ b/pkg/gitreceive/storage/slug_builder_info.go
@@ -1,0 +1,28 @@
+package storage
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/deis/builder/pkg/gitreceive/git"
+)
+
+type SlugBuilderInfo struct {
+	PushKey string
+	PushURL string
+	TarKey  string
+	TarURL  string
+}
+
+func NewSlugBuilderInfo(s3Client *s3.S3, appName, slugName string, gitSha *git.SHA) *SlugBuilderInfo {
+	tarKey := fmt.Sprintf("home/%s/tar", slugName)
+	// this is where workflow tells slugrunner to download the slug from, so we have to tell slugbuilder to upload it to here
+	pushKey := fmt.Sprintf("home/%s/push", fmt.Sprintf("%s:git-%s", appName, gitSha.Short))
+
+	return &SlugBuilderInfo{
+		PushKey: pushKey,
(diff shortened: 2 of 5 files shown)


pkg/gitreceive/storage/slug_builder_info.go
```diff
@@ -0,0 +1,28 @@
...
+	PushKey string
+	PushURL string
+	TarKey  string
+	TarURL  string
+}
+
+func NewSlugBuilderInfo(s3Client *s3.S3, appName, slugName string, gitSha *git.SHA) *SlugBuilderInfo {
+	tarKey := fmt.Sprintf("home/%s/tar", slugName)
+	// this is where workflow tells slugrunner to download the slug from, so we have to tell slugbuilder to upload it to here
+	pushKey := fmt.Sprintf("home/%s/push", fmt.Sprintf("%s:git-%s", appName, gitSha.Short))
```
krancour: The double `fmt.Sprintf` makes this harder to read.

Consider this:

```
pushKey := fmt.Sprintf("home/%s:git-%s/push", appName, gitSha.Short)
```

arschles: good call, thanks

krancour: Also, shouldn't this be the full sha?  That seems to be what workflow currently tells the slugrunner to use.

pkg/gitreceive/git/sha.go
```diff
@@ -0,0 +1,31 @@
...
+	return fmt.Sprintf("git sha %s was too short", e.sha)
+}
+
+type SHA struct {
+	Full  string
+	Short string
+}
+
+func NewSha(rawSha string) (*SHA, error) {
+	if len(rawSha) < fullShaLen {
```
krancour: Why are we only concerned with the case where the sha is too short?  Wouldn't being too long or containing invalid characters be equally problematic?

Here's a regex: `^[\da-f]{40}$`

pkg/gitreceive/git/sha.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+func (e ErrGitShaTooShort) Error() string {
+	return fmt.Sprintf("git sha %s was too short", e.sha)
+}
+
+type SHA struct {
+	Full  string
+	Short string
+}
```
krancour: I'm probably overthinking this, but assuming a `SHA` is meant to be immutable, should these fields maybe no be exported and have exported accessors instead?

pkg/gitreceive/storage/slug_builder_info.go
```diff
@@ -0,0 +1,29 @@
...
+	PushURL string
+	TarKey  string
+	TarURL  string
+}
+
+// NewSlugBuilderInfo creates and populates a new SlugBuilderInfo based on the given data
+func NewSlugBuilderInfo(s3Endpoint, appName, slugName string, gitSha *git.SHA) *SlugBuilderInfo {
+	tarKey := fmt.Sprintf("home/%s/tar", slugName)
+	// this is where workflow tells slugrunner to download the slug from, so we have to tell slugbuilder to upload it to here
+	pushKey := fmt.Sprintf("home/%s:git-%s/push", appName, gitSha.Short)
```
krancour: Workflow is actually telling the slugrunner to use the full SHA.

pkg/gitreceive/git/sha.go
arschles: You're right, the regex is the ideal solution

pkg/gitreceive/storage/slug_builder_info.go
arschles: Yes, full seems to be working. Fixed in https://github.com/arschles/builder/commit/1e4f08f0eae7c0efecefa4047b3ecd8fb5af72b4

pkg/gitreceive/git/sha.go
arschles: Fixed in https://github.com/arschles/builder/commit/0ff588721161a854c4496d733602b493fdf3f49f

pkg/gitreceive/storage/slug_builder_info.go
arschles: Fixed in https://github.com/arschles/builder/commit/1e4f08f0eae7c0efecefa4047b3ecd8fb5af72b4

pkg/gitreceive/git/sha.go
arschles: IMO you're not overthinking, and yes, it's meant to be immutable. It probably makes sense to make these and the fields in `SlugBuilderInfo` non-exported and build exported read-only accessors (as much as I loathe that pattern) as you said, but also so that it's clearer that users should use the `New...` functions. Commit forthcoming

arschles: Fixed in https://github.com/deis/builder/commit/2ed89afd5b49ac20b72964b2c9183f4420107c0e

pkg/gitreceive/git/sha.go
```diff
@@ -0,0 +1,37 @@
...
+	return fmt.Sprintf("git sha %s was invalid", e.sha)
+}
+
+type SHA struct {
+	full  string
+	short string
+}
+
+func NewSha(rawSha string) (*SHA, error) {
+	if !shaRegex.Match([]byte(rawSha)) {
```
krancour: `shaRegex.MatchString(rawSha)` can be used to avoid the cast to `[]byte` since those bytes aren't used anywhere else.

krancour: But that's a nit.

krancour:
LGMT from a perspective of all the refactoring you have done and the fix on URL formatting.  I'm still not sure that this will address the apparent problem of built slugs not being uploaded at all to object storage, but the root cause of that isn't actually clear to me.

arschles:
@krancour thanks for looking. The root problem was that the push URL didn't have `/git/...` in it, and also wasn't using the full git sha (both fixed in https://github.com/arschles/builder/commit/1e4f08f0eae7c0efecefa4047b3ecd8fb5af72b4). The rest of this PR involved making the necessary refactors to be able to add regression tests for the change, and also writing those tests. This patch has lots of changes, but IMO they're in the name of better testing of the git-receive hook and still scoped tightly.

pkg/gitreceive/git/sha.go
arschles: Sure, I'll make the change. It should clean up the code here

adamreese:
Still a WIP?

arschles:
Nope, just updated the title

pkg/gitreceive/git/sha.go
```diff
@@ -0,0 +1,37 @@
...
+func (e ErrInvalidGitSha) Error() string {
+	return fmt.Sprintf("git sha %s was invalid", e.sha)
+}
+
+type SHA struct {
+	full  string
+	short string
+}
+
+func NewSha(rawSha string) (*SHA, error) {
```
bacongobbler: i know go vet doesn't like this sort of stuff (package name in function). Mind changing this to `.New()` instead?

arschles: Hm, it doesn't seem to complain (see below), but I agree - change forthcoming.

```console
ENG000656:builder aaronschlesinger$ go vet $(glide nv)
ENG000656:builder aaronschlesinger$ 
```

pkg/gitreceive/git/sha.go
```diff
@@ -0,0 +1,37 @@
...
+	if !shaRegex.MatchString(rawSha) {
+		return nil, ErrInvalidGitSha{sha: rawSha}
+	}
+	return &SHA{full: rawSha, short: rawSha[0:8]}, nil
+}
+
+func (s SHA) Full() string  { return s.full }
```
bacongobbler: We cold also provide a `.String()` func that's an alias here so we don't have to constantly use `.Full()` or `.Short()`, depending on which one makes more sense as the string representation.

pkg/gitreceive/git/sha.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"fmt"
+	"regexp"
+)
+
+const (
+	// this constant represents the length of a shortened git sha - 8 characters long
+	shortShaIdx = 8
+	fullShaLen  = 40
```
bacongobbler: As far as I can see this isn't actually used. Was there intention to use this somewhere?

pkg/gitreceive/git/sha.go
arschles: @bacongobbler I purposefully didn't provide a `String()` since it's ambiguous - I intended for this structure to "force" clients of this `struct` to decide whether they want full or shortened SHAs

pkg/gitreceive/git/sha.go
bacongobbler: Ah, I remember what my knowledge is coming from. In deisctl we had a function called `.FleetClient()` in package `client/fleet`. go vet didn't like it due to stuttering and so it preferred `.New` instead. That function was created before we introduced go vet though and we never got around to fixing it.

pkg/gitreceive/git/sha.go
bacongobbler: makes sense

pkg/gitreceive/git/sha.go
arschles: @bacongobbler ha! I was just typing a comment that took back what I said - I forgot that the package name was `git` herhe package name here is `git`, so there's no conflict

pkg/gitreceive/git/sha.go
arschles: You're right. It was a relic left over after a refactor.

pkg/gitreceive/git/sha.go
bacongobbler: lapse of (short term) memory. I thought the package name was `sha`. Carry on!

pkg/gitreceive/git/sha.go
arschles: Fixed in https://github.com/deis/builder/commit/8e8c0a6d7c14997aaf919fe92170dd4f613a718d

smothiki:
I mean about the builder like there is a new package for storage, configs, for deploying pods . So made changes accordingly

Issue #142, opened 2016-02-02
change the embedded dockerbuilder image to the official deis one


arschles:
requires https://github.com/deis/dockerbuilder/issues/4

smothiki:
refer https://github.com/deis/dockerbuilder/pull/6

arschles:
@smothiki is this done?

PR #147, opened 2016-02-05
fix(builder): change dockerfile builder image

smothiki:
fixes https://github.com/deis/builder/issues/142

### DIFF
--- a/pkg/gitreceive/k8s_util.go
+++ b/pkg/gitreceive/k8s_util.go
@@ -12,11 +12,10 @@ import (
 )
 
 const (
-	slugBuilderName   = "deis-slugbuilder"
-	slugBuilderImage  = "quay.io/deisci/slugbuilder:v2-beta"
-	dockerBuilderName = "deis-dockerbuilder"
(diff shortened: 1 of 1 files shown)


pkg/gitreceive/k8s_util.go
```diff
@@ -12,11 +12,10 @@ import (
 )
 
 const (
-	slugBuilderName   = "deis-slugbuilder"
-	slugBuilderImage  = "quay.io/deisci/slugbuilder:v2-beta"
-	dockerBuilderName = "deis-dockerbuilder"
-	// change the value back to quay.io/deisci/dockerbuilder:v2-beta
-	dockerBuilderImage = "smothiki/dockerbuilder:v1.8"
+	slugBuilderName    = "deis-slugbuilder"
+	slugBuilderImage   = "quay.io/deisci/slugbuilder:v2-beta"
```
mboersma: This image path should be configurable ideally, not hard-coded. Let's fix that later.

smothiki: sure . I'm thinking if setting environment variables in builder RC for images.

PR #148, opened 2016-02-05
ref(builder): remove etcd and confd

### LINKED ISSUE #81: [Meta] Replace etcd usage with something else
This should be done after #72. See the table below for analysis

Place it's used | Suggested solution | Status | Idea
------------------- | -------------------------- | -------- | ------
`confd` regenerates the `authorized_keys` file | instead of relying on confd to regenerate it, just call to workflow to get keys on every push | Not Done. Work tracked in #122 | Have builder make an API call to the controller to check the validity of a key for the given app. Tracked in #122 
The git-receive hook relies on the `/deis/controller/builderKey` etcd key | This key holds auth information for builder to talk to workflow. Instead of storing it in etcd, store in a k8s secret. See deis/workflow#245, which is doing this work | Done in #72 | N/A
`rootfs/bin/cleanup` | This removes application directories as applications are deleted (and their deletion is published via etcd) in the controller API | Not Done | Have the controller periodically query the k8s API for apps. [deis/router](https://github.com/deis/router) does something similar.

This issue should remove confd as well.

### LINKED ISSUE #122: make builder read keys from database
Builder reads keys from etcd. We should avoid doing that and read keys from database

aledbf:
closes #122
closes #81

~~Requires deis/workflow#352~~ (merged)


aledbf:
ping @arschles

aledbf:
To test https://github.com/deis/workflow/pull/352 is required

krancour:
I'm very highly in favor of this, however, it looks like there are commits in this PR that aren't directly part of the narrow concern of removing etcd and confd.  It's _all_ good stuff, but should this be broken up into a few separate PRs for easier review and traceability?

pkg/commands.go
```diff
@@ -3,27 +3,12 @@ package pkg
...
-//  -message (string): The message to log when entering sleep.
-func Sleep(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
-	dur := p.Get("duration", 10*time.Millisecond).(time.Duration)
-	msg := p.Get("messages", "Sleeping").(string)
-	log.Info(c, msg)
-	time.Sleep(dur)
-	log.Info(c, "Woke up.")
-	return true, nil
-}
-
```
arschles: I'm *very* happy that this is being removed!

:100:

pkg/confd/confd.go
```diff
@@ -1,96 +0,0 @@
...
-		if err := cmd.Wait(); err != nil {
-			// If confd exits, builder will stop functioning as intended. So
-			// we stop builder and let the environment restart.
-			log.Errf(c, "Stopping builder. confd exited with error: %s", err)
-			os.Exit(37)
-		}
-	})
-
-	return true, nil
-}
```
arschles: Same here regarding removal :+1:

pkg/etcd/etcd.go
```diff
@@ -1,535 +0,0 @@
...
-		return nil
-	case 500:
-		time.Sleep(retrySleep)
-		return nil
-	case 200:
-		return nil
-	default:
-		return fmt.Errorf("Unhandled HTTP Error: %s %d", last.Status, last.StatusCode)
-	}
-}
```
arschles: And here!

aledbf:
> but should this be broken up into a few separate PRs for easier review and traceability?

The changes in the code removes dependencies in packages (etcd and confd), the use of docker, multiples packages now are not required (Dockerfile), tests that makes no sense (interaction with workflow to read app configuration for instance), etc. which make this a difficult PR to break a part. I can remove the "replace current log with github.com/deis/pkg/log commit" but that makes no sense since I already need to update the dependencies for this PR.

That said if this PR is not accepted in the current form please indicate the way to open multiples PR that depends one on another.

pkg/gitreceive/run.go
```diff
@@ -3,18 +3,15 @@ package gitreceive
...
 	"os"
 	"strings"
 
-	"github.com/deis/builder/pkg/gitreceive/log"
 	"github.com/deis/builder/pkg/gitreceive/storage"
+	"github.com/deis/pkg/log"
 
-	client "k8s.io/kubernetes/pkg/client/unversioned"
-)
+	builder_conf "github.com/deis/builder/pkg/conf"
```
arschles: Can you avoid using `_` in the package name and rename this `builderconf`?

pkg/gitreceive/storage/bucket.go
```diff
@@ -26,6 +28,8 @@ func BucketExists(svc *s3.S3, bucketName string) (bool, error) {
 	return true, nil
 }
 
+// CreateBucket creates a new bucket in the S3 API compatible storage or
+// return an error in case the bucket exists
```
arschles: Thanks for all these docs

pkg/sshd/server.go
```diff
@@ -8,7 +8,7 @@
 package sshd
 
 import (
-	"encoding/binary"
+	// "encoding/binary"
```
arschles: can this be removed?

pkg/sshd/sshd.go
```diff
@@ -178,35 +124,55 @@ func Configure(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrup
...
-
-	i, j := 0, 0
-	for ; i < len(buf); i++ {
-		if i > 0 && i%2 == 0 {
-			fp[j] = ':'
-			j++
-		}
-		fp[j] = buf[i]
-		j++
+func userInfoFromKey(key string) (*UserInfo, error) {
```
arschles: Can you put this `func`, along with `controllerURLStr` into a `github.com/deis/pkg/controller` package?

pkg/sshd/sshd.go
```diff
@@ -178,35 +124,55 @@ func Configure(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrup
...
+	ret := &UserInfo{}
+	if err := json.NewDecoder(res.Body).Decode(ret); err != nil {
+		return nil, err
+	}
+	return ret, nil
+}
+
+func controllerURLStr(additionalPath ...string) string {
+	host := os.Getenv("DEIS_WORKFLOW_SERVICE_HOST")
+	port := os.Getenv("DEIS_WORKFLOW_SERVICE_PORT")
```
arschles: if either of these are empty, we should return an error

arschles:
@aledbf reviewed and left some comments. Also, I agree with @krancour, it would be *very* helpful if you could split some things out into different pull requests. Below is a list of some ideas for things to split out, each into separate PRs.

- The transition from using `github.com/deis/builder/pkg/gitreceive/log` to `github.com/deis/pkg/log`
- The change from `tarUrl` => `tarURL` and `putUrl` => `putURL` in `slugBuildCase`
- The restructuring and switch to `alpine:3.3` in the `Dockerfile`. The removal of `etcdctl` and `confd` should obviously stay in this PR

Thanks again for this change!

glide.lock
```diff
@@ -55,14 +50,19 @@ imports:
...
   version: 4239b77079c7b5d1243b7b4736304ce8ddb6f0f2
+  repo: https://github.com/armon/go-radix
 - name: github.com/armon/gomdb
   version: 151f2e08ef45cb0e57d694b2562f351955dff572
+  repo: https://github.com/armon/gomdb
+- name: github.com/arschles/assert
+  version: 6882f85ccdc7c1822b146d1a6b0c2c48f91b5140
 - name: github.com/aws/aws-sdk-go
-  version: 87b1e60a50b09e4812dee560b33a238f67305804
+  version: ""
```
arschles: you may have to run `make glideup` to get this version back

arschles:
@aledbf the code LGTM to me, but I haven't tested it yet. I also removed `closes #151` from the description since that change is gone from this PR (thanks, by the way, for removing it!)

aledbf:
@arschles done

arschles:
thanks @aledbf

pkg/controller/utils.go
```diff
@@ -0,0 +1,79 @@
...
+		return "", fmt.Errorf("missing required '%v' environment variable", portEnvName)
+	}
+
+	return fmt.Sprintf("http://%s:%s/%s", host, port, strings.Join(additionalPath, "/")), nil
+}
+
+func UserInfoFromKey(key string) (*UserInfo, error) {
+	keyB64 := base64.RawURLEncoding.EncodeToString([]byte(key))
+	url, err := controllerURLStr("v2", "hooks", "key", keyB64)
+	if err != nil {
```
smothiki: @helgi is this the correct url endpoint ```/v2/hooks/key/key-base64``` to get user info ? Just wanna cross check.
I think its the ```/v2/keys``` calls keyview set

helgi: Yeah it's `key`, I dropped the plural so the old hooks worked still under `keys` (the one we will remove). It also made sense since we were querying on a singular key instead of getting many keys back.

technosophos:
Over the long run we should tighten up security on the controller endpoint, but that's a task for a different day.

I am :100: behind this one. I am so glad to see etcd and confd pulled out. I also did a very careful read through the SSH handling, and I think this method is good.

technosophos:
And thanks, @aledbf. This was a huge PR to work through.

arschles:
@aledbf I'm testing this now

arschles:
@aledbf I tested with a deis/workflow#352 build and a build of this branch. The `git push` failed, with this in the builder logs:

```console
[info] Checking closer.
[info] Accepted connection.
[debug] Starting ssh authentication
[debug] Checking auth for user key ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDIqRuNwhjntdJWuAVykr/873X8zzKo6Ms1Vx70BQx0wir8TpaLZEY6CqqKDrMbHZ2Z0ZV2ZITs9fC81GqIdmmDFXZxNfx+B1lSR3ZpmZpQpprtSCevZXihIgy+ND50Hp/wk+3VU54FxhudIlJgpPjb/o5vQFhiyM3ynR5gH3slWVaq9C0TkgXCnTzukGzSTeL7wYPNmLomkrAS0nk0yRfoUZcwmD++HMgEmYlhTbnMlkB3nxzEf/JQxhY6xCHrbtNbRkINCY21dHrsrr/MvBvD8FoKnKpxHX2+HNXZbe7Xl87L9o1OuXrtR1crvq+r+1fPjaynGir07zr9mgJPxouPL/e4ppxTL//vt1kVkWWkh/B+GyXEmP38bQGYMpEA7cAndlxPlOki35JYwDNn5CENQpDp8F4+JsKIzAF1zmkBIA7ngg0cSHHilNgwZXmX7h+7nngLgFIpP8h7A9fCpAKhHUfFUj+Zgl9Xm44+sZOwVBnVijK326TgVDFTUXjE/Xwny+3ERgYwBfOwOKmusNFnS0XHbmh+qa/+D8qge5bKilq48pKHzwngM/U6OwMxmSXTuHclLLen3Ime30TOiPzAhokrVNz/Z3VAkfBuJHby68SAKUgczUEU81wz5wFEt1n1sIJ5V49KMRGaSWb+eWvW81yA7NkDSjnsMLa/IF/ADQ==
[error] Failed handshake: EOF (&{{0xc820226770}})
```

And this in the workflow logs:

```console
10.168.1.1 "GET /v2/hooks/key/c3NoLXJzYSBBQUFBQjNOemFDMXljMkVBQUFBREFRQUJBQUFDQVFESXFSdU53aGpudGRKV3VBVnlrci84NzNYOHp6S282TXMxVng3MEJReDB3aXI4VHBhTFpFWTZDcXFLRHJNYkhaMlowWlYyWklUczlmQzgxR3FJZG1tREZYWnhOZngrQjFsU1IzWnBtWnBRcHBydFNDZXZaWGloSWd5K05ENTBIcC93ayszVlU1NEZ4aHVkSWxKZ3BQamIvbzV2UUZoaXlNM3luUjVnSDNzbFdWYXE5QzBUa2dYQ25UenVrR3pTVGVMN3dZUE5tTG9ta3JBUzBuazB5UmZvVVpjd21EKytITWdFbVlsaFRibk1sa0Izbnh6RWYvSlF4aFk2eENIcmJ0TmJSa0lOQ1kyMWRIcnNyci9NdkJ2RDhGb0tuS3B4SFgyK0hOWFpiZTdYbDg3TDlvMU91WHJ0UjFjcnZxK3IrMWZQamF5bkdpcjA3enI5bWdKUHhvdVBML2U0cHB4VEwvL3Z0MWtWa1dXa2gvQitHeVhFbVAzOGJRR1lNcEVBN2NBbmRseFBsT2tpMzVKWXdETm41Q0VOUXBEcDhGNCtKc0tJekFGMXpta0JJQTduZ2cwY1NISGlsTmd3WlhtWDdoKzdubmdMZ0ZJcFA4aDdBOWZDcEFLaEhVZkZVaitaZ2w5WG00NCtzWk93VkJuVmlqSzMyNlRnVkRGVFVYakUvWHdueSszRVJnWXdCZk93T0ttdXNORm5TMFhIYm1oK3FhLytEOHFnZTViS2lscTQ4cEtIenduZ00vVTZPd014bVNYVHVIY2xMTGVuM0ltZTMwVE9pUHpBaG9rclZOei9aM1ZBa2ZCdUpIYnk2OFNBS1VnY3pVRVU4MXd6NXdGRXQxbjFzSUo1VjQ5S01SR2FTV2IrZVd2VzgxeUE3TmtEU2puc01MYS9JRi9BRFE9PQo HTTP/1.1" 400 - "deis-builder"
```

Any ideas on what went wrong?

aledbf:
@arschles yes

> Any ideas on what went wrong?

yes, here https://github.com/deis/builder/pull/148/files#diff-7549650024dfe83fbb2b4fdf653c2d49R43
workflow is using https://docs.python.org/3/library/base64.html#base64.urlsafe_b64decode
and the error in python is:
```
old-mbp:builder aledbf$ python
Python 2.7.10 (default, Aug 22 2015, 20:33:39)
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.59.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import base64
>>> base64.urlsafe_b64decode("c3NoLXJzYSBBQUFBQjNOemFDMXljMkVBQUFBREFRQUJBQUFDQVFESXFSdU53aGpudGRKV3VBVnlrci84NzNYOHp6S282TXMxVng3MEJReDB3aXI4VHBhTFpFWTZDcXFLRHJNYkhaMlowWlYyWklUczlmQzgxR3FJZG1tREZYWnhOZngrQjFsU1IzWnBtWnBRcHBydFNDZXZaWGloSWd5K05ENTBIcC93ayszVlU1NEZ4aHVkSWxKZ3BQamIvbzV2UUZoaXlNM3luUjVnSDNzbFdWYXE5QzBUa2dYQ25UenVrR3pTVGVMN3dZUE5tTG9ta3JBUzBuazB5UmZvVVpjd21EKytITWdFbVlsaFRibk1sa0Izbnh6RWYvSlF4aFk2eENIcmJ0TmJSa0lOQ1kyMWRIcnNyci9NdkJ2RDhGb0tuS3B4SFgyK0hOWFpiZTdYbDg3TDlvMU91WHJ0UjFjcnZxK3IrMWZQamF5bkdpcjA3enI5bWdKUHhvdVBML2U0cHB4VEwvL3Z0MWtWa1dXa2gvQitHeVhFbVAzOGJRR1lNcEVBN2NBbmRseFBsT2tpMzVKWXdETm41Q0VOUXBEcDhGNCtKc0tJekFGMXpta0JJQTduZ2cwY1NISGlsTmd3WlhtWDdoKzdubmdMZ0ZJcFA4aDdBOWZDcEFLaEhVZkZVaitaZ2w5WG00NCtzWk93VkJuVmlqSzMyNlRnVkRGVFVYakUvWHdueSszRVJnWXdCZk93T0ttdXNORm5TMFhIYm1oK3FhLytEOHFnZTViS2lscTQ4cEtIenduZ00vVTZPd014bVNYVHVIY2xMTGVuM0ltZTMwVE9pUHpBaG9rclZOei9aM1ZBa2ZCdUpIYnk2OFNBS1VnY3pVRVU4MXd6NXdGRXQxbjFzSUo1VjQ5S01SR2FTV2IrZVd2VzgxeUE3TmtEU2puc01MYS9JRi9BRFE9PQo".strip()).
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/base64.py", line 112, in urlsafe_b64decode
    return b64decode(s, '-_')
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/base64.py", line 76, in b64decode
    raise TypeError(msg)
TypeError: Incorrect padding
```

arschles:
@aledbf I suspected after talking with @helgi. I'm gonna try using `base64.StdEncoding` and `base64.URLEncoding` to see if either works, as both are un-padded. Will report back.

technosophos:
Can we just put the key in the body and not base64 encode it?

arschles:
@technosophos yes, we could do that, but I'd like to keep this as a standard `GET` request, compliant with good REST design. I've opened https://github.com/aledbf/builder/pull/1 (to merge into the branch represented here) to switch to using a fingerprint in the URL path. deis/workflow#352 has been merged, so aledbf/builder#1 will need to be merged into this branch before this can be merged.

Using a build of aledbf/builder#1 against the standard `deis-dev` helm chart, I verified that the `git push` succeeds.

technosophos:
Packing a huge amount of data into the URL is not good design.

arschles:
It _was_ a huge amount of data, but now it a standard public key fingerprint. I wouldn't consider that a huge amount of data, and certainly doesn't come close to any path length restrictions I know of.

technosophos:
I am definitely okay with the fingerprint, provided @helgi does not object.

helgi:
Already implemented and merged into workflow using fingerprint

arschles:
:100: @helgi

pkg/controller/utils.go
```diff
@@ -0,0 +1,103 @@
...
+const (
+	hostEnvName = "DEIS_WORKFLOW_SERVICE_HOST"
+	portEnvName = "DEIS_WORKFLOW_SERVICE_PORT"
+)
+
+// UserInfo represent the required information from a user to make a push and interact with deis/workflow
+type UserInfo struct {
+	Username    string
+	Key         string
+	FingerPrint string
```
mboersma: This should probably be `Fingerprint` without InterCaps (like `Username`). Not worth holding up this PR though.

arschles: @mboersma can you file an issue for that?

arschles:
@aledbf FYI I forgot to mark as `LGTM2` before I merged, but it was definitely fine beforehand ;)

PR #149, opened 2016-02-08
fix(*): add readiness & liveness probes

### LINKED ISSUE #89: bring manifests up to date with deis/charts
in particular, the replication controller manifest needs some work

https://github.com/deis/charts/blob/master/deis-dev/manifests/deis-builder-rc.yaml

### LINKED ISSUE #92: Add healthcheck and liveness check to builder manifest
The probes should check:

- The ssh server is running
- The s3 server is accessible
- The k8s API is accessible

When this is done, do https://github.com/deis/charts/issues/87

Ref deis/deis#4809

arschles:
Fixes #92 
Fixes #89 
Ref deis/deis#4809
When this is done, merge deis/charts#93

# PR Overview

- Added a health check server with a single endpoint (`/healthz`), intended to be used by the `livenessProbe` and `readinessProbe`.
  - This endpoint checks the accessibility of the Kubernetes API & the object storage system, and checks that the SSH server is ready to serve
  - The health check server utilizes two `interface`s - for S3 interaction and Kubernetes API interaction - so that those two components can be mocked and the server can be more easily unit tested
  - The handler for the `/healthz` endpoint does all three checks concurrently, and only fails if one or more of them don't complete before a hard-coded timeout, or any of them fail before the timeout
- Added a `(github.com/deis/builder/pkg/sshd).Circuit` type, which acts as a simple, concurrency-safe on/off switch for communication across concurrency barriers. A `Circuit` is used to allow the SSH server to tell the health check server when it's ready to serve
- Modified `boot.go` to serve the SSH server and the health check server concurrently, but to kill the process with an error code if either fails
- Modified the manifests to closely resemble those in github.com/deis/charts (#89)
- Added liveness and readiness checks to the `deis-builder-rc.yaml` manifest
- Modified the `Makefile` to only look for non-vendored packages when `make test` is run, which makes all runs faster

# Still TODO

- [x] Check k8s server on probe request
- [x] Check that the server is actually running on probe request
- [x] Tests for `healthZHandler`
- [x] `TestOpenCloseConcurrent`
- [x] `make test` output: `pkg/sshd/server_test.go:106: not enough arguments in call to Serve`

### DIFF
--- a/boot.go
+++ b/boot.go
@@ -9,6 +9,7 @@ import (
 	"github.com/deis/builder/pkg"
 	"github.com/deis/builder/pkg/conf"
 	"github.com/deis/builder/pkg/gitreceive"
+	"github.com/deis/builder/pkg/gitreceive/storage"
 	"github.com/deis/builder/pkg/sshd"
 	pkglog "github.com/deis/pkg/log"
 )
@@ -42,6 +43,13 @@ func main() {
 					pkglog.Err("getting config for %s [%s]", serverConfAppName, err)
 					os.Exit(1)
 				}
+				s3Client, err := storage.GetClient
+				if err != nil {
+					pkglog.Err("getting s3 client [%s]", err)
+					os.Exit(1)
+				}
+				pkglog.Info("starting health check server on port %d", cnf.HealthSrvPort)
+				go healthsrv.Start(cnf.HealtHSrvPort, s3Client)
 				pkglog.Info("starting SSH server on %s:%d", cnf.SSHHostIP, cnf.SSHHostPort)
 				os.Exit(pkg.Run(cnf.SSHHostIP, cnf.SSHHostPort, "boot"))
 			},
--- a/boot.go
+++ b/boot.go
@@ -12,7 +12,11 @@ import (
 	"github.com/deis/builder/pkg/gitreceive/storage"
 	"github.com/deis/builder/pkg/healthsrv"
 	"github.com/deis/builder/pkg/sshd"
+<<<<<<< 032d8fd56928af3492a8449e226d36d5324b8d2c
 	pkglog "github.com/deis/pkg/log"
+=======
+	kcl "k8s.io/kubernetes/pkg/client/unversioned"
+>>>>>>> fix(boot.go,pkg/healthsrv): add kubernetes API checks in the healthz endpoint
 )
 
 const (
@@ -49,10 +53,15 @@ func main() {
 					pkglog.Err("getting s3 client [%s]", err)
 					os.Exit(1)
 				}
+				kubeClient, err := kcl.NewInCluster()
+				if err != nil {
+					pkglog.Err("getting kubernetes client [%s]", err)
+					os.Exit(1)
+				}
 				pkglog.Info("starting health check server on port %d", cnf.HealthSrvPort)
 				healthSrvCh := make(chan error)
 				go func() {
-					if err := healthsrv.Start(cnf.HealthSrvPort, s3Client); err != nil {
+					if err := healthsrv.Start(cnf.HealthSrvPort, kubeClient.Namespaces(), s3Client); err != nil {
 						healthSrvCh <- err
 					}
 				}()
--- a/boot.go
+++ b/boot.go
@@ -30,8 +30,8 @@ func main() {
 	if os.Getenv("DEBUG") == "true" {
 		pkglog.DefaultLogger.SetDebug(true)
 		cookoolog.Level = cookoolog.LogDebug
+		log.Printf("Running in debug mode")
 	}
-	pkglog.Debug("Running in debug mode")
 
 	app := cli.NewApp()
 
--- a/boot.go
+++ b/boot.go
@@ -10,6 +10,7 @@ import (
 	"github.com/deis/builder/pkg/conf"
 	"github.com/deis/builder/pkg/gitreceive"
 	"github.com/deis/builder/pkg/gitreceive/storage"
+	"github.com/deis/builder/pkg/healthsrv"
 	"github.com/deis/builder/pkg/sshd"
 	pkglog "github.com/deis/pkg/log"
 )
@@ -43,15 +44,33 @@ func main() {
 					pkglog.Err("getting config for %s [%s]", serverConfAppName, err)
 					os.Exit(1)
 				}
-				s3Client, err := storage.GetClient
+				s3Client, err := storage.GetClient(cnf.HealthSrvTestStorageRegion)
 				if err != nil {
 					pkglog.Err("getting s3 client [%s]", err)
 					os.Exit(1)
 				}
 				pkglog.Info("starting health check server on port %d", cnf.HealthSrvPort)
-				go healthsrv.Start(cnf.HealtHSrvPort, s3Client)
+				healthSrvCh := make(chan error)
+				go func() {
+					if err := healthsrv.Start(cnf.HealthSrvPort, s3Client); err != nil {
+						healthSrvCh <- err
+					}
+				}()
+
 				pkglog.Info("starting SSH server on %s:%d", cnf.SSHHostIP, cnf.SSHHostPort)
-				os.Exit(pkg.Run(cnf.SSHHostIP, cnf.SSHHostPort, "boot"))
+				sshCh := make(chan int)
+				go func() {
+					sshCh <- pkg.Run(cnf.SSHHostIP, cnf.SSHHostPort, "boot")
+				}()
+
+				select {
+				case err := <-healthSrvCh:
+					pkglog.Err("Error running health server (%s)", err)
+					os.Exit(1)
+				case i := <-sshCh:
+					pkglog.Err("Unexpected SSH server stop with code %d", i)
+					os.Exit(i)
+				}
 			},
 		},
 		{
--- a/boot.go
+++ b/boot.go
@@ -45,6 +45,8 @@ func main() {
 					pkglog.Err("getting config for %s [%s]", serverConfAppName, err)
 					os.Exit(1)
 				}
+				circ := sshd.NewCircuit()
+
 				s3Client, err := storage.GetClient(cnf.HealthSrvTestStorageRegion)
 				if err != nil {
 					pkglog.Err("getting s3 client [%s]", err)
@@ -58,7 +60,7 @@ func main() {
 				pkglog.Info("starting health check server on port %d", cnf.HealthSrvPort)
 				healthSrvCh := make(chan error)
 				go func() {
-					if err := healthsrv.Start(cnf.HealthSrvPort, kubeClient.Namespaces(), s3Client); err != nil {
+					if err := healthsrv.Start(cnf.HealthSrvPort, kubeClient.Namespaces(), s3Client, circ); err != nil {
 						healthSrvCh <- err
 					}
 				}()
@@ -66,7 +68,7 @@ func main() {
 				pkglog.Info("starting SSH server on %s:%d", cnf.SSHHostIP, cnf.SSHHostPort)
 				sshCh := make(chan int)
 				go func() {
-					sshCh <- pkg.Run(cnf.SSHHostIP, cnf.SSHHostPort, "boot")
+					sshCh <- pkg.RunBuilder(cnf.SSHHostIP, cnf.SSHHostPort, circ)
 				}()
 
 				select {
--- a/boot.go
+++ b/boot.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"log"
 	"os"
 	"runtime"
 
@@ -49,15 +50,15 @@ func main() {
 
 				s3Client, err := storage.GetClient(cnf.HealthSrvTestStorageRegion)
 				if err != nil {
-					pkglog.Err("getting s3 client [%s]", err)
+					log.Printf("Error getting s3 client (%s)", err)
 					os.Exit(1)
 				}
 				kubeClient, err := kcl.NewInCluster()
 				if err != nil {
-					pkglog.Err("getting kubernetes client [%s]", err)
+					log.Printf("Error getting kubernetes client [%s]", err)
 					os.Exit(1)
 				}
-				pkglog.Info("starting health check server on port %d", cnf.HealthSrvPort)
+				log.Printf("Starting health check server on port %d", cnf.HealthSrvPort)
 				healthSrvCh := make(chan error)
 				go func() {
 					if err := healthsrv.Start(cnf.HealthSrvPort, kubeClient.Namespaces(), s3Client, circ); err != nil {
@@ -65,7 +66,7 @@ func main() {
 					}
 				}()
 
-				pkglog.Info("starting SSH server on %s:%d", cnf.SSHHostIP, cnf.SSHHostPort)
+				log.Printf("Starting SSH server on %s:%d", cnf.SSHHostIP, cnf.SSHHostPort)
 				sshCh := make(chan int)
 				go func() {
 					sshCh <- pkg.RunBuilder(cnf.SSHHostIP, cnf.SSHHostPort, circ)
@@ -73,10 +74,10 @@ func main() {
 
 				select {
 				case err := <-healthSrvCh:
-					pkglog.Err("Error running health server (%s)", err)
+					log.Printf("Error running health server (%s)", err)
 					os.Exit(1)
 				case i := <-sshCh:
-					pkglog.Err("Unexpected SSH server stop with code %d", i)
+					log.Printf("Unexpected SSH server stop with code %d", i)
 					os.Exit(i)
 				}
 			},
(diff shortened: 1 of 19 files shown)


pkg/healthsrv/healthz_handler.go
```diff
@@ -0,0 +1,73 @@
...
+			log.Err(str)
+			http.Error(w, str, http.StatusServiceUnavailable)
+			return
+		}
+		var rsp healthZResp
+		for _, buck := range lbOut.Buckets {
+			rsp.S3Buckets = append(rsp.S3Buckets, convertBucket(buck))
+		}
+
+		nsList, err := nsLister.List(labels.Everything(), fields.Everything())
```
aledbf: Why not `framework.NewInformer` like [here](https://github.com/kubernetes/contrib/blob/master/service-loadbalancer/service_loadbalancer.go#L569) to not hit the apiserver unnecessarily

arschles: I'm unfamiliar with `k8s.io/kubernetes/pkg/controller/framework`, but it looks like it does *not* hit the API server, which I do need to do. Did I misunderstand something?

Also, if you know of a cheaper operation than listing namespaces (I'm assuming the common case is that there will not be many), I'd love to hear it!

aledbf: @arschles I will post a snippet for listing namespaces using `controller.framework`. The idea behind this abstraction is that uses watch and an implicit cache that reduces the round trips to api-server and transfer size

arschles: That would be great, thanks. I'd definitely like to reduce general load on the k8s master.

krancour:
Is this still WIP?  The WIP in the title and the "awaiting review" tag are giving me mixed signals.

arschles:
@krancour sorry about that. Yes, still in progress. I changed the label to indicate so.

arschles:
Tested myself. To test yourself, spin up a deis-dev cluster and `kd delete rc deis-builder && kd create -f manifests/deis-builder-rc.yaml && kd get pod -w`. Once you see the pod come up, wait for it to go into `Running` and `1/1

arschles:
@krancour this is now ready for review

pkg/sshd/circuit.go
```diff
@@ -0,0 +1,56 @@
...
+// Circuit is a concurrency-safe data structure that can take one of two states at any point in time:
+//
+// - OpenState - non functional
+// - ClosedState - functional
+//
+//
+type Circuit struct {
+	bit uint32
+}
+
```
smothiki: why not byte or bool?

arschles: see below - the `atomic` package has no facilities for manipulating either one of those types

smothiki:
why not modify fetcher to health serv. Do we need another http server to handle health serv. Already fetcher has a health check handler

pkg/sshd/circuit.go
smothiki: Okay I didn't know that trying to learn

smothiki:
Anyways we are thinking of moving away from fetcher . So please Ignore my previous comment

pkg/sshd/circuit.go
arschles: basically, I'm using `atomic` here because the circuit can "get away" with simple bit flipping - and using atomic operations can easily do that in a cheaper way than with a mutex

boot.go
```diff
@@ -43,10 +47,51 @@ func main() {
...
+				kubeClient, err := kcl.NewInCluster()
+				if err != nil {
+					log.Printf("Error getting kubernetes client [%s]", err)
+					os.Exit(1)
+				}
+				log.Printf("Starting health check server on port %d", cnf.HealthSrvPort)
+				healthSrvCh := make(chan error)
+				go func() {
+					if err := healthsrv.Start(cnf.HealthSrvPort, kubeClient.Namespaces(), s3Client, circ); err != nil {
+						healthSrvCh <- err
```
smothiki: Do we have to check if namepsace is present or not ?  builder won't start if namespace is not present or k8s cluster is not working

arschles: Well, the goal here is to check that the k8s API is accessible. Given that the builder has started, we can assume that the namespace under which it runs (also specified in `POD_NAMESPACE`) exists. The call that the health server makes is to just get a list of namespaces, and the goal there is only to see if the master is accessible.

I think @aledbf had an idea for a cheaper method of determining whether the API server was accessible, but if you have ideas I'd like to hear them too.

smothiki:
If k8s API is not accessible how come it's builder fault . Also it not making sense to me to check if API is accessible or not as builder wont start in the first place. 
I agree checking S3 storage is accessible or not 
This is overall amazing code. I feel its more complex design 
If we are coming up a different health server why do we need an entire class to atomically set a value. We can just write one  more handler in the health server that sets the circuit state value and ssh.server.go calls that handler before this https://github.com/deis/builder/blob/master/pkg/sshd/server.go#L73 and health check handle polls change of circuit state till the timeout is over.

smothiki:
Also liveness probes shouldn't be checking S3 API accessibility

boot.go
aledbf: @arschles if the goal is just check the status of API server why not just run:
```
	res := c.Get().AbsPath("/healthz").Do()
	if res.Error() != nil{
		// not possible to reach the server
	}
```

smothiki:
Already server handles ping request, but ping method wasn't implemented https://github.com/deis/builder/blob/master/pkg/sshd/server.go#L202. I think liveness probe should call ping

arschles:
>If k8s API is not accessible how come it's builder fault

It's not the builder's fault, but the builder shouldn't be able to accept any request to do work when object storage, a dependency without which it can't do its job, is unavailable to it.

>If we are coming up a different health server why do we need an entire class to atomically set a value

We don't necessarily need this `Circuit` type, but it presents the simplest abstraction between the SSH server and the health check server I could find. The basic requirement is that the SSH server can communicate its status to the health server in a concurrency safe manner.

I tried a few other implementations, including a `sync.WaitGroup` implementation and a channel based implementation, but both ended up being more code with more complexity.

>We can just write one more handler in the health server that sets the circuit state value and ssh.server.go calls that handler before this https://github.com/deis/builder/blob/master/pkg/sshd/server.go#L73

I'm not 100% sure I understand what you mean here, but the closure of the circuit happens right above that line [here](https://github.com/arschles/deis-builder/blob/probes/pkg/sshd/server.go#L73), and that is the logic that effectively communicates to the health check server that the ssh server is ready

IMO It doesn't make as much sense to call an HTTP endpoint than to make a function call from within the same process.

>health check handle polls change of circuit state till the timeout is over.

The health check server essentially does poll the circuit state because Kubernetes polls the health check  endpoint. I opted not to implement any polling logic in the health check endpoint in favor of just reusing the Kubernetes logic.

>Already server handles ping request, but ping method wasn't implemented https://github.com/deis/builder/blob/master/pkg/sshd/server.go#L202. I think liveness probe should call ping

I think implementing the health check logic in the SSH ping handler is suboptimal. Here's my thinking:

- The readiness and liveness probes could become expensive in the future as the SSH server grows in complexity. In fact, they may add more CPU load onto the existing server (I have no data on this, purely a guess)
- The health check logic would be unnecessarily mixed with the SSH logic, violating separation of concerns. This point also relates to the previous
- The end-user expects to do an ssh ping operation to determine whether the server up and ready for their `git push`es. By putting health check logic as-is into the ping (and I believe the health check logic we have now is 100% necessary), we allow any end-user of the builder to (possibly unwillingly or unknowingly) exercise our object store and the Kubernetes API arbitrarily.

boot.go
arschles: @aledbf that's much better - thanks!

arschles: @aledbf here's the rendered error when that code runs:

`couldn't get version/kind; json parse error: invalid character 'o' looking for beginning of value`

aledbf: > Objections? 

None

arschles: #180

manifests/deis-builder-rc.yaml
```diff
@@ -15,24 +16,41 @@ spec:
...
       containers:
         - name: deis-builder
-          imagePullPolicy: Always
           image: quay.io/deisci/builder:v2-beta
+          imagePullPolicy: Always
           ports:
             - containerPort: 2223
+              name: ssh
             - containerPort: 3000
+              name: fetcher
```
bacongobbler: Looks like we missed a spot in #171 :)

manifests/deis-builder-rc.yaml
```diff
@@ -40,10 +58,6 @@ spec:
             - name: builder-key-auth
               mountPath: /var/run/secrets/api/auth
               readOnly: true
-            # not currently running minio with SSL support. see https://github.com/deis/minio/pull/22 for more detail
```
bacongobbler: would you mind just removing this code entirely for now?

pkg/builder.go
```diff
@@ -28,7 +27,7 @@ const (
 // Git.
 //
 // Run returns on of the Status* status code constants.
-func Run(sshHostIP string, sshHostPort int, cmd string) int {
+func RunBuilder(sshHostIP string, sshHostPort int, sshServerCircuit *sshd.Circuit) int {
```
bacongobbler: is there a reason we're renaming this from `builder.Run` to `builder.RunBuilder`? Not too sure how this relates to healthchecks

arschles: I removed `cmd` and changed the name to add specificity to this function. Objections to leaving it in?

For context, `cmd`'s purpose was to tell `Run` what cookoo route to run, but none were supported except for `"boot"` (which corresponds to running the builder SSH server).

manifests/deis-builder-rc.yaml
arschles: Thanks - removed in 958e0b1

manifests/deis-builder-rc.yaml
arschles: yep, it's removed here ;)

pkg/builder.go
arschles: Also, the reason I had to touch this function in the first place was to add the `sshServerCircuit` parameter. That's the part I missed in my last comment, and that's what related to health checks.

pkg/sshd/circuit.go
```diff
@@ -0,0 +1,61 @@
+package sshd
+
+import (
+	"fmt"
+	"sync/atomic"
+)
+
+// CircuitState represents the state of a Circuit
+type CircuitState uint32
```
bacongobbler: Maybe I'm not understanding the lower level synchronization types here, but from reading up to understand this bit of code from sync/atomic's docs:

> These functions require great care to be used correctly. Except for special, low-level applications, synchronization is better done with channels or the facilities of the sync package. Share memory by communicating; don't communicate by sharing memory.

Not saying this should be refactored (this looks good to me), but I'm curious on your decision to go this route instead of using a combination of a boolean and mutal exclusion locks instead. Looking at the code, it looks like we only need two circuit states: open and closed.

pkg/builder.go
bacongobbler: Gotcha. No objections here, just curious :)

manifests/deis-builder-rc.yaml
bacongobbler: oh I meant like removing the commented-out code entirely rather than leaving it as a comment

pkg/sshd/circuit.go
arschles: Yup, good question.

A `bool`/`sync.RWMutex` or channel-based implementation would work just as well here, and both would fit fine into this abstraction. I went the `sync/atomic` route mainly because there's no need to protect a _critical section_ of logic - just the mutation of a single boolean - and `sync/atomic` can do concurrency-safe bit-twiddling well.

Basically, I originally checked this code in thinking it would be simplest and _not_ because I had any performance or specific `sync/atomic` benefits in mind. Now that I've gotten a few questions on it, I wonder if it'd be best to refactor it using a mutex, which seems to be more familiar. Thoughts?

_Here's some more on `sync/atomic` if you care:_

Atomic integer and pointer manipulation usually does lend itself to lower level systems problems like databases, etc..., but the core concept still applies here - atomically manipulating bits. We're just approximating that single bit as a `uint32`.

Also, on many systems, the `sync/atomic` approach is more performant in a few different regards, but that's getting into scheduler and OS level implementation details, which is where I get fuzzy anyway.

manifests/deis-builder-rc.yaml
arschles: Yup, it's removed entirely. [Here](https://github.com/arschles/deis-builder/blob/probes/manifests/deis-builder-rc.yaml) is the file on my branch, which is represented in the PR - notice no mention of SSL

pkg/builder.go
arschles: Cool. Thanks for letting me slip this sorta-not-related change in ;)

arschles:
@bacongobbler are you referring to the `AbsPath("/healthz")` bug?

manifests/deis-builder-rc.yaml
bacongobbler: wow. I was **really** derping hard there. Carry on.

arschles: it happens ;)

bacongobbler:
Sorry, that comment was in regards to refactoring out pkg/log to use the standard log package as in [this commit](https://github.com/arschles/deis-builder/commit/5a209b6b3f1a79e666cfae3e9e8ca3c088143c56). Github doesn't really show well that I was making a comment on that commit.

arschles:
Ah, got it. I see the tiny text where it says you commented on the commit now.

Filed https://github.com/deis/pkg/issues/19 to track

pkg/sshd/circuit.go
bacongobbler: > Now that I've gotten a few questions on it, I wonder if it'd be best to refactor it using a mutex, which seems to be more familiar. Thoughts?

Honestly it's all bikeshedding, which I *hate*. I'm quite happy with this implementation, and it serves its purpose as required, so it looks good to me. :)

On the plus side, I just learned about a new package today! :tada:

arschles: Alrighty, thanks. Enjoy your new atomicity?

aledbf:
Code LGTM

arschles:
Gonna hold off on merging this until the discussion at https://github.com/deis/builder/pull/149#issuecomment-183126063 is complete

technosophos:
FWIW, here is my view on health checking logic:

- The readiness check should ONLY pass if the underlying system believes itself to be ready to successfully handle a request.
- A liveness probe should ONLY pass if the system is in a condition where it could successfully handle a request.

Neither probes are about establishing "fault", the purpose of the probes is to indicate that the system believes itself to be in a position to fulfill its service contract. That's it. If it can't reach a database, cache, API endpoint, etc., it should fail its check.

That said, I am a little concerned with probes being dependent on the state of Kubernetes itself. So the k8s API service availability is sort of a gray area. Originally, I was opposed to failing healthz based on k8s API server problems, but I've changed my mind. Here's the use case that caused me to change my mind:

- Builder is working just fine.
- A well-meaning admin tries to lock down the API server by enforcing stricter security policies
- Builder can no longer hit the API server.

In this case, builder SHOULD fail because it can no longer fulfill its service contract. If a policy was changed, k8s itself will restart the pod and it will come up successful. Otherwise, it will go into crash backoff, which is exactly what ought to happen.

helgi:
A liveness probe should *not* fail if k8s is down or s3 is reachable and here is why:

A liveness probe will cause a pod to *restart* when those external components fail, which in turn means the user can not even do `git push` but instead gets a unreachable hostname if there are no builders to service the request.

As a user (developer) I want to do a `git push` which fails *gracefully* if dependant services are down with a good error message, along the lines of  "unable to process the git push right now, X and Y are down".

Builder is *alive* and can give good information back to the user, making the liveness probe depend on 3rd party things is a *bad* UX since the whole builder becomes unavailable at that point. I'm proposing we do the deeper health check during `git push` and `readiness` probe but not `liveness` probe. Different purposes and different external informational needs by devs / operators.

I'd like to hear the opposite side on how we'd handle the scenario of builder being down and telling the developers (and operators) what is going on and how they would effectively debug things when the builder is effectively left in a "running but not ready" state until s3 / k8s / etc comes up?

bacongobbler:
To chime in my thoughts as well:

> In this case, builder SHOULD fail because it can no longer fulfill its service contract. If a policy was changed, k8s itself will restart the pod and it will come up successful. Otherwise, it will go into crash backoff, which is exactly what ought to happen.

This is great for an administrator, but not for a user. In the current implementation the user has no view into why the builder is down. If we were to go down this route, a status page (similar to status pages like http://status.docker.com/) would be invaluable to a user to see what systems are down, bridging the communication gap between users and administrators.

> I'd like to hear the opposite side on how we'd handle the scenario of builder being down and telling the developers (and operators) what is going on and how they would effectively debug things when the builder is effectively left in a "running but not ready" state until s3 / k8s / etc comes up?

^^

Another proposed solution offline was to gracefully fail a `git push`, but internally perform health checks on upstream services like k8s/minio, which would then be logged if they're reporting as down. That way we can pick that up in https://github.com/deis/monitor with alerting (not implemented in telegraf AFAIK but not something that we couldn't do ourselves)

arschles:
Ok, after extensive internal discussion, consensus is as follows:

- For now, and probably through the `v2-beta` release, the logic in this PR will suffice as-is
- I will be creating issues for the following enhancements, to be done at various points in the release cycle
  - Have health probes check the status of the controller
  - Have health probes ping the local SSH server (instead of what they do now, which is only verify that it has started properly)
  - Have the `git push` logic respond with an informative error message during downstream dependency failures

PR #161, opened 2016-02-10
fix(pkg/gitreceive/build.go): send the entire slug URL to the controller

arschles:
The controller now expects builder to send the *entire* slug URL to it. It then forwards that URL to the slugrunner that it launches.

This makes builder compatible with the changes in https://github.com/deis/workflow/pull/346

### DIFF
--- a/pkg/gitreceive/build.go
+++ b/pkg/gitreceive/build.go
@@ -217,6 +217,8 @@ func build(conf *Config, s3Client *s3.S3, kubeClient *client.Client, builderKey,
 	}
 	if !usingDockerfile {
 		buildHook.Dockerfile = ""
+		// need this to tell the controller what URL to give the slug runner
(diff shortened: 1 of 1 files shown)


smothiki:
tested now 
```
spec:
  containers:
  - args:
    - start
    - web
    env:
    - name: PORT
      value: "5000"
    - name: SLUG_URL
      value: http://10.0.119.169:9000/git/home/rugged-aqualung:git-e1590887/push/slug.tgz
    - name: DEBUG
      value: "1"
    - name: DEIS_APP
      value: rugged-aqualung
    - name: DEIS_RELEASE
      value: v2
    - name: DOCKERIMAGE
      value: "1"
    image: quay.io/deisci/slugrunner:v2-beta
    imagePullPolicy: Always
    livenessProbe:
      httpGet:
        path: /
        port: 5000
```

PR #154, opened 2016-02-09
ref(rootfs/Dockerfile): upgrade to alpine 3.3, clean up the dockerfile (WIP)

arschles:
Reduces the number of layers that will rarely be updated.

Fixes #153 
Closes #120

PR #120, opened 2016-01-21
fix(Dockerfile): upgrade to alpine 3.3 (WIP)

arschles:

PR #165, opened 2016-02-10
ref(rootfs/Dockerfile): upgrade to alpine 3.3, clean up the dockerfile

aledbf:
closes #154

### DIFF
--- a/rootfs/Dockerfile
+++ b/rootfs/Dockerfile
@@ -1,8 +1,7 @@
-FROM alpine:3.1
+FROM alpine:3.3
 
 # install common packages
 RUN apk add --update-cache \
-  curl \
   bash \
   sudo \
   git \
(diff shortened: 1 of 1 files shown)


arschles:
@aledbf testing this now

rootfs/Dockerfile
```diff
@@ -1,4 +1,14 @@
-FROM alpine:3.1
+FROM alpine:3.3
+
+# install git and configure gituser
+ENV GITHOME /home/git
+ENV GITUSER git
+# this is so the minio client (https://github.com/minio/mc) works properly
+ENV DOCKERIMAGE=1
+ENV DEIS_RELEASE 2.0.0-dev
```
bacongobbler: This envvar will be modified on every release. We should probably punt this to the bottom of the Dockerfile.

arschles:
closes #120

arschles:
@aledbf I think we can remove the `curl` installation from the Dockerfile, no?

arschles:
@aledbf a manual test of `git push` seemed to work fine (except that the resulting slugrunner failed to run, which I believe was solved in #161). LGTM

rootfs/Dockerfile
aledbf: > This envvar will be modified on every release. We should probably punt this to the bottom of the Dockerfile.

That's correct but I would prefer a rebuild of the image if a variable is changed and not use the cache of the previous layers

bacongobbler: What if no changed occurred that release? The version bump would be a no-op and the cached layers would stay the same, which would make the diff between two versions minimal. What would be the reason behind busting the cache in this situation?

aledbf: > What if no changed occurred that release

Maybe not in the `go` code but if a new version of openssl is available (to fix one vulnerability) the image will not reflect this change because of the cached layers.

That said, please tell me what to do with the order of the variables

bacongobbler: If there's a critical security issue we can purge the docker cache in Travis and force a rebuild. That way we can fix it in master before we cut a new release. :)

Let's move it to the bottom

PR #166, opened 2016-02-10
fix(builder): if exit code from build pod is not 0 it must return an error

aledbf:
This avoid the creation of new releases that will end in a `CrashLoopBackoff` status

### DIFF
--- a/pkg/gitreceive/build.go
+++ b/pkg/gitreceive/build.go
@@ -190,7 +190,24 @@ func build(conf *Config, s3Client *s3.S3, kubeClient *client.Client, builderKey,
 	if err != nil {
 		return fmt.Errorf("fetching builder logs (%s)", err)
 	}
-	log.Debug("size of logs streamed %v", size)
+	log.Debug("size of streamed logs %v", size)
+
+	// check the state and exit code of the build pod.
+	// if the code is not 0 return error
+	if err := waitForPodEnd(kubeClient, newPod.Namespace, newPod.Name, conf.BuilderPodTickDuration(), conf.BuilderPodWaitDuration()); err != nil {
+		return fmt.Errorf("error getting builder pod status (%s)", err)
+	}
+	buildPod, err := kubeClient.Pods(newPod.Namespace).Get(newPod.Name)
+	if err != nil {
+		return fmt.Errorf("error getting builder pod status (%s)", err)
+	}
+
+	for _, containerStatus := range buildPod.Status.ContainerStatuses {
+		state := containerStatus.State.Terminated
+		if state.ExitCode != 0 {
+			return fmt.Errorf("Stopping build.")
+		}
+	}
 
 	// poll the s3 server to ensure the slug exists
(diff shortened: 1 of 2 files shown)


arschles:
@aledbf testing this now

aledbf:
@arschles to test this you use something like this:
```
mkdir slugfailure
cd slugfailure
git init
echo "web: true" > Procfile
git add Procfile && git commit -m "Import 1"
deis create slugfailure && git push deis master
```

arschles:
@aledbf the test you described works fine, as well as a similar test using https://github.com/deis/example-go

pkg/gitreceive/build.go
```diff
@@ -190,7 +190,21 @@ func build(conf *Config, s3Client *s3.S3, kubeClient *client.Client, builderKey,
...
+	log.Debug("size of streamed logs %v", size)
+
+	// check the state and exit code of the build pod.
+	// if the code is not 0 return error
+	if err := waitForPodEnd(kubeClient, newPod.Namespace, newPod.Name, conf.BuilderPodTickDuration(), conf.BuilderPodWaitDuration()); err != nil {
+		return fmt.Errorf("error getting builder pod status (%s)", err)
+	}
+	buildPod, err := kubeClient.Pods(newPod.Namespace).Get(newPod.Name)
+	if err != nil {
+		return fmt.Errorf("error getting builder pod status")
```
bacongobbler: we should be returning the error here as well so it isn't swallowed up

aledbf: oops, sorry about that. Fixed.

pkg/gitreceive/build.go
```diff
@@ -190,7 +190,21 @@ func build(conf *Config, s3Client *s3.S3, kubeClient *client.Client, builderKey,
...
+	// check the state and exit code of the build pod.
+	// if the code is not 0 return error
+	if err := waitForPodEnd(kubeClient, newPod.Namespace, newPod.Name, conf.BuilderPodTickDuration(), conf.BuilderPodWaitDuration()); err != nil {
+		return fmt.Errorf("error getting builder pod status (%s)", err)
+	}
+	buildPod, err := kubeClient.Pods(newPod.Namespace).Get(newPod.Name)
+	if err != nil {
+		return fmt.Errorf("error getting builder pod status (%s)", err)
+	}
+	state := buildPod.Status.ContainerStatuses[0].State.Terminated
```
bacongobbler: I feel like we should be checking all container statuses, not just the first one in the pod. @arschles, thoughts?

aledbf: @bacongobbler `waitForPodEnd` waits until the build pod [ends](https://github.com/deis/builder/pull/166/files#diff-d9098478974874a718d578a2f5912a77R164) so the state only can be Terminated

aledbf: @bacongobbler I didn't understand your comment. ContainerStatuses[0] is because the builder pod only contains 1 container. I can change that to avoid a future update to this code.

bacongobbler: oh right. this is just for the build pod, not for the final pod deployed. Disregard

PR #170, opened 2016-02-11
bug(builder): avoid simultaneous git push

### LINKED ISSUE #168: [Question]: Avoid simultaneous git push for the same application
In V1 this is not possible. I'm not sure if this should be allowed or not.
Running two simultaneous `git push` I see this in the builder logs:
```
Waiting for git-receive to run.
Waiting for deploy.
[info] Deploy complete.
[warning] Unreported error: error: Ref refs/heads/master is at 2bd4702d7ede8c417a0a070755e12b99ddbba6b7 but expected 0000000000000000000000000000000000000000
[info] Deploy complete.
```

aledbf:
```
$ git push deis master
fatal: remote error: Another git push is ongoing. Aborting...
```

closes #168

### DIFF
--- a/pkg/sshd/lock.go
+++ b/pkg/sshd/lock.go
@@ -1,0 +1,68 @@
+package sshd
+
+import (
+	"fmt"
+	"sync"
+	"time"
+)
+
+// RepositoryLock interface that allows the creation of a lock associated
+// with a repository name to avoid simultaneous git operations
+type RepositoryLock interface {
+	// Lock aquires a lock for a repository. In the case the repository is already locked
+	// it waits until a timeout to get the lock. If it was not possible to get the
+	// lock after the timeout an error is returned
+	Lock(repoName string, timeout time.Duration) error
+	// Unlock releases the lock for a repository or returns an error if the specified
+	// name doesn't exist. In the case the repository is already locked it waits until
+	// a timeout to get the lock. If it was not possible to get the lock after the timeout
+	// an error is returned
+	Unlock(repoName string, timeout time.Duration) error
+}
+
+// NewInMemoryRepositoryLock returns a new instance of a RepositoryLock
+func NewInMemoryRepositoryLock() RepositoryLock {
+	return &inMemoryRepoLock{
+		mutex:   &sync.RWMutex{},
+		dataMap: make(map[string]bool),
+	}
+}
+
+type inMemoryRepoLock struct {
+	mutex   *sync.RWMutex
+	dataMap map[string]bool
+}
+
+// Lock aquires a lock associated with the specified name.
+// This implementation do not uses the timeout
+func (rl *inMemoryRepoLock) Lock(repoName string, timeout time.Duration) error {
+	rl.mutex.Lock()
+	defer rl.mutex.Unlock()
+
+	_, exists := rl.dataMap[repoName]
+	if !exists {
+		rl.dataMap[repoName] = true
+		return nil
+	}
+
+	return fmt.Errorf("repository %q already locked", repoName)
+}
+
+// Unlock releases the lock for a repository or returns
+// an error if the specified name doesn't exist.
+// This implementation do not uses the timeout
+func (rl *inMemoryRepoLock) Unlock(repoName string, timeout time.Duration) error {
+	rl.mutex.Lock()
+	defer rl.mutex.Unlock()
+
+	locked, exists := rl.dataMap[repoName]
+	if !exists {
+		return fmt.Errorf("repository %q not found", repoName)
+	}
+
+	if locked {
+		delete(rl.dataMap, repoName)
+	}
+
+	return nil
+}
--- a/pkg/sshd/lock_test.go
+++ b/pkg/sshd/lock_test.go
@@ -34,7 +34,7 @@ func TestInvalidUnlock(t *testing.T) {
 	key := "fakeid"
 	err := rl.Unlock(key, time.Duration(0))
 	if err == nil {
-		t.Fatalf("expected error but returned nil", err)
+		t.Fatal("expected error but returned nil")
 	}
 }
 
--- a/pkg/sshd/lock_test.go
+++ b/pkg/sshd/lock_test.go
@@ -1,0 +1,85 @@
+package sshd
+
+import (
+	"testing"
+	"time"
+)
+
+const (
+	callbackTimeout = 1 * time.Second
+)
+
+func TestSingleLock(t *testing.T) {
+	rl := NewInMemoryRepositoryLock()
+	key := "fakeid"
+	callbackCh := make(chan interface{})
+	go lockAndCallback(rl, key, callbackCh)
+	verifyCallbackHappens(t, callbackCh)
+}
+
+func TestSingleLockUnlock(t *testing.T) {
+	rl := NewInMemoryRepositoryLock()
+	key := "fakeid"
+	callbackCh := make(chan interface{})
+	go lockAndCallback(rl, key, callbackCh)
+	verifyCallbackHappens(t, callbackCh)
+	err := rl.Unlock(key, time.Duration(0))
+	if err != nil {
+		t.Fatalf("unexpected error %v", err)
+	}
+}
+
+func TestInvalidUnlock(t *testing.T) {
+	rl := NewInMemoryRepositoryLock()
+	key := "fakeid"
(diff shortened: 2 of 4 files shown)


bacongobbler:
We should probably add unit tests for RepositoryLock

aledbf:
@bacongobbler unit tests added

pkg/sshd/lock.go
```diff
@@ -0,0 +1,67 @@
...
+
+// NewRepositoryLock returns a new instance of a key mutex.
+func NewRepositoryLock() RepositoryLock {
+	return &repoLock{
+		mutexMap: make(map[string]bool),
+	}
+}
+
+type repoLock struct {
+	sync.RWMutex
```
arschles: can you give this a field name

pkg/sshd/lock.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// NewRepositoryLock returns a new instance of a key mutex.
+func NewRepositoryLock() RepositoryLock {
+	return &repoLock{
+		mutexMap: make(map[string]bool),
+	}
+}
+
+type repoLock struct {
```
arschles: can you call this `inMemoryRepoLock`

pkg/sshd/lock.go
```diff
@@ -0,0 +1,67 @@
...
+type RepositoryLock interface {
+	// Lock aquires a lock for a repository
+	Lock(repoName string) error
+	// Unlock releases the lock for a repository or returns
+	// an error if the specified name doesn't exist.
+	Unlock(repoName string) error
+}
+
+// NewRepositoryLock returns a new instance of a key mutex.
+func NewRepositoryLock() RepositoryLock {
```
arschles: can you rename this `NewInMemoryRepositoryLock`?

pkg/sshd/lock.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+type repoLock struct {
+	sync.RWMutex
+	mutexMap map[string]bool
+}
+
+// Lock aquires a lock associated with the specified name
+func (rl *repoLock) Lock(repoName string) error {
+	rl.RWMutex.Lock()
```
arschles: it should be enough to `RLock` this if it's just for accessing the map, no?

pkg/sshd/lock.go
```diff
@@ -0,0 +1,67 @@
...
+// NewRepositoryLock returns a new instance of a key mutex.
+func NewRepositoryLock() RepositoryLock {
+	return &repoLock{
+		mutexMap: make(map[string]bool),
+	}
+}
+
+type repoLock struct {
+	sync.RWMutex
+	mutexMap map[string]bool
```
arschles: shouldn't this map from `string` to `*sync.Mutex`, so that you can lock each repo individually?

rootfs/Dockerfile
```diff
@@ -11,6 +11,10 @@ RUN apk add --update-cache \
   xz \
   && rm -rf /var/cache/apk/*
 
+# install git and configure gituser
+ENV GITHOME /home/git
+ENV GITUSER git
```
arschles: :+1: - I ran into this in my PR as well

pkg/sshd/lock.go
aledbf: no because that produces a data race (I tried your suggestion using `go test -race -cover lock.go lock_test.go`)

arschles: @aledbf I think that's because you're changing the `bool` values of `mutexMap`. See my comment at https://github.com/deis/builder/pull/170#discussion_r52618236 regarding changing the `bool` to a `*sync.Mutex`

pkg/sshd/lock.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"fmt"
+	"sync"
+)
+
+// RepositoryLock interface that allows the creation of a lock associated
+// with a repository name to avoid simultaneous git operations
+type RepositoryLock interface {
+	// Lock aquires a lock for a repository
+	Lock(repoName string) error
```
arschles: @aledbf can this and `Unlock` below also accept a `timeout time.Duration`, so that thie future implementations can wait to acquire the lock

aledbf: @arschles I tried that but it feels awkward see a git push just waiting. The behavior in this PR provides the same UX than in deis V1

arschles: @aledbf I think for now we could solve that by passing `time.Duration(0)` to the `Lock` func.

If we use a network-based lock in the future, though, we're going to need to account for network delays, failure cases, etc... and a timeout is a simple first solution for doing so.

arschles: Also, @aledbf - if / when we do decide to use a timeout greater than 0, we could still provide a progress indicator in the console - I wouldn't be opposed to that.

arschles:
@aledbf how did you test this?

pkg/sshd/lock.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// Lock aquires a lock associated with the specified name
+func (rl *inMemoryRepoLock) Lock(repoName string, timeout time.Duration) error {
+	rl.mutex.Lock()
+	defer rl.mutex.Unlock()
+
+	_, exists := rl.dataMap[repoName]
+	if !exists {
+		fmt.Printf("Creating lock for repository %v\n", repoName)
```
arschles: I don't think this should log here

pkg/sshd/lock.go
```diff
@@ -0,0 +1,63 @@
...
+// with a repository name to avoid simultaneous git operations
+type RepositoryLock interface {
+	// Lock aquires a lock for a repository
+	Lock(repoName string, timeout time.Duration) error
+	// Unlock releases the lock for a repository or returns
+	// an error if the specified name doesn't exist.
+	Unlock(repoName string, timeout time.Duration) error
+}
+
+// NewInMemoryRepositoryLock returns a new instance of a key mutex.
```
arschles: These docs don't seem to be current. What is a "key mutex". Also, it appears that the timeouts aren't respected here - can you add some docs to indicate that fact?

aledbf:
@arschles executing simultaneous `git push` in the same app and `git push` in the middle of the build.

pkg/sshd/lock.go
arschles: Also, is it by design that the in memory mutex doesn't respect the timeout?

arschles:
ok, thanks

arschles:
also, would it be possible to add some tests for locking in `pkg/sshd/server_test.go`?

pkg/sshd/lock.go
bacongobbler: or at the very least we could keep this, but punt it to debug level

arschles:
@aledbf tested and this works for me as well. let me know when you'd like another review

pkg/sshd/server.go
```diff
@@ -215,6 +223,19 @@ func (s *server) answer(channel ssh.Channel, requests <-chan *ssh.Request, sshCo
 					req.Reply(ok, nil)
 					break
 				}
+
+				repoName := parts[1]
```
arschles: can you ensure that `len(parts) >= 2`. I know that it should in the common case, but it would be good to be defensive

pkg/sshd/lock_test.go
```diff
@@ -0,0 +1,85 @@
...
+		t.Fatalf("unexpected error %v", err)
+	}
+}
+
+func TestInvalidUnlock(t *testing.T) {
+	rl := NewInMemoryRepositoryLock()
+	key := "fakeid"
+	err := rl.Unlock(key, time.Duration(0))
+	if err == nil {
+		t.Fatalf("expected error but returned nil", err)
```
arschles: this fails `go vet`: `no formatting directive in Fatalf call` I think this can remove the `err` (since it's nil) and make this a `t.Fatal`

pkg/sshd/server.go
aledbf: @arschles that was already present here https://github.com/deis/builder/pull/170#discussion-diff-53344595R221

arschles: k. sorry, I should have read up

arschles:
@aledbf thanks for adding the additional test. I'm marking LGTM pending CI passing

PR #171, opened 2016-02-11
ref(builder): remove fetcher

### LINKED ISSUE #6: Remove the Fetcher
The buildpack builder has a [fetcher](https://github.com/deis/builder/tree/master/bpbuilder/fetcher), which acts as a slug and tar specific network storage system. Its implementation is to store slugs and tars in specific locations on the local filesystem. It appears that the fetcher's goal is to be a fallback in case S3 is not configured for the install.

We're adding support for minio (see `Object Storage` in https://github.com/deis/deis/issues/4743), which is a more general purpose object storage system that can run in the cluster (currently its implementation is also to store data on the filesystem) in v2 Alpha.

I'm proposing that we remove the fetcher and require that either S3 or Minio access is configured for builder and [slugbuilder](http://github.com/deis/slugbuilder).

cc/ @smothiki

aledbf:
closes #139
closes #6


arschles:
@aledbf is this tested?

aledbf:
@arschles yes but I forgot to push the `Dockerfile` change required to compile

arschles:
@aledbf can you also submit a PR on https://github.com/deis/charts that removes the fetcher related things in the deis-builder manifests? I'm guessing the only change will have to be on the `deis-builder-rc.yaml` manifest

arschles:
@aledbf can you also remove the `DEIS_BUILDER_SERVER_FETCHER_PORT` environment variable from the Dockerfile?

bacongobbler:
good to merge :)

PR #173, opened 2016-02-11
fix(.travis.yml): run the docker-build target on all branches

### LINKED ISSUE #169: Make the CI system run `docker build` even on branches

arschles:
This PR allows us to test Dockerfile changes in PRs

Fixes #169

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ DEV_ENV_CMD := ${DEV_ENV_PREFIX} ${DEV_ENV_IMAGE}
 
 # SemVer with build information is defined in the SemVer 2 spec, but Docker
 # doesn't allow +, so we use -.
-VERSION ?= git-$(shell git rev-parse --short HEAD)
+VERSION ?= git$(shell git rev-parse --short HEAD)
 BINARY_DEST_DIR := rootfs/usr/bin
 # Common flags passed into Go's linker.
 LDFLAGS := "-s -X main.version=${VERSION}"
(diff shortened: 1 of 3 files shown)


arschles:
Note: https://github.com/arschles/deis-builder/commit/ca94ea26ce2f85ee3c96c09d202e56ff6d4ec872 was made because it fixes an issue in the Dockerfile, which this PR caught

Ironic Image

Makefile
```diff
@@ -20,7 +20,7 @@ IMAGE_PREFIX ?= deis
 # Docker Root FS
 BINDIR := ./rootfs
 
-DEIS_REGISTRY ?= ${DEV_REGISTRY}/
+DEIS_REGISTRY ?= quay.io/
```
bacongobbler: typo?

arschles: Nope. `${DEV_REGISTRY}` doesn't exist in this makefile.

bacongobbler: It might be better to specify $DEV_REGISTRY as an environment variable in .travis.yml instead. That way we don't have hardcoded registries in the Makefile. DEV_REGISTRY is a holdover from v1 :)

bacongobbler: Or DEIS_REGISTRY, w/e tickles your fancy

arschles: got it. ok, will do

bacongobbler: For reference: https://github.com/deis/prototype-repo/blob/master/Makefile#L26

arschles: ah, right. thanks!

PR #186, opened 2016-02-18
fix(boot.go,pkg/*): implement a deleted app cleaner

### LINKED ISSUE #167: Delete local copy of git repos when apps are deleted
Previously, app deletion was done by removing a key from etcd. Builder would watch app keys, see the change and delete the local copy. Now that https://github.com/deis/builder/issues/81 is complete, the system is no longer in place and local copies are never removed. So this needs to be fixed.

Possibilities:

- Builder runs a "reaper" goroutine in builder that iterates the git home directory every `$INTERVAL`, asks the controller if the app still exists, and deletes the local directory if it does
- The controller calls a builder API to tell it when an app is deleted. When builder receives this API call, it deletes the local folder. This system will not work well when we support more than one builder

arschles:
# Overview

This PR introduces a background goroutine that polls the Kubernetes API for the list of namespaces and compares those names with those in its local git folder. The cleaner then deletes all directories that exist locally but not in the namespace list.

# Testing Instructions

To test, follow these steps:

1. Create an app and `git push` to it from repository A
2. `deis apps:destroy -a your_app`
3. Wait at least the polling interval (the default is 1 second)
4. Re-create an app of the same name from repository B (so that the `deis` tool creates the same remote in repo B). Make sure repository B has a different set of refs from A
5. `git push` from repo B

The push should be successful and the app should run properly.

Still TODO:

- [x] Lock repositories in-memory while they're being deleted. The SSH server should reject pushes to repos that are currently being deleted. Depends on #170 for the `RepositoryLock` primitive in order to achieve this locking
- [x] Make the cleaner's poll duration configurable (it's currently hard-coded)

Notes:

- This PR also introduces global locks on `git push`es, effectively removing some of the functionality in #170, but still preventing concurrent git pushes. We may want to hold on this PR in order to get back to full feature parity with that introduced in #170
- Assuming the previous were fixed, this PR could still introduce an edge case that can impact performance negatively. See https://github.com/deis/builder/issues/192 for details and a proposed solution.
- There's a polling window for the cleaner. If a user deletes an app, re-creates it, and pushes a different repo within that window, they'll see errors. See https://github.com/deis/builder/issues/193 for details and a proposed solution.

Fixes #167

### DIFF
--- a/pkg/cleaner/cleaner.go
+++ b/pkg/cleaner/cleaner.go
@@ -64,6 +64,8 @@ func stripSuffixes(strs []string, suffix string) []string {
 		idx := strings.LastIndex(str, suffix)
 		if idx >= 0 {
 			ret[i] = str[:idx]
+		} else {
+			ret[i] = str
 		}
 	}
 	return ret
--- a/pkg/cleaner/cleaner.go
+++ b/pkg/cleaner/cleaner.go
@@ -3,6 +3,7 @@ package cleaner
 
 import (
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
@@ -10,7 +11,6 @@ import (
 
 	"github.com/deis/builder/pkg/k8s"
 	"github.com/deis/builder/pkg/sshd"
-	"github.com/deis/pkg/log"
 	"k8s.io/kubernetes/pkg/api"
 	"k8s.io/kubernetes/pkg/fields"
 	"k8s.io/kubernetes/pkg/labels"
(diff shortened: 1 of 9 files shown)


pkg/cleaner/cleaner.go
```diff
@@ -0,0 +1,90 @@
...
+
+// getDiff gets the directories that are not in namespaceList
+func getDiff(namespaceList []api.Namespace, dirs []string) []string {
+	var ret []string
+
+	// create a set of lowercase namespace names
+	namespacesSet := make(map[string]struct{})
+	for _, ns := range namespaceList {
+		lowerName := strings.ToLower(ns.Name)
+		namespacesSet[lowerName] = struct{}{}
```
smothiki: This only works if namespaces are not case sensitive

arschles: well, are we case sensitive on app name? cc/ @helgi

helgi: I do not believe we attempt to make it case insensitive but I don't know for sure. We do not do any lowercasing anywhere specifically

smothiki: My bad namespaces are lowercase [a-z0-9]([-a-z0-9]*[a-z0-9])?)

arschles: ok, we'll have to make sure we keep this naming standard intact ;)

helgi: Tracked down, `[a-z0-9-]+` is the Application Regex, so that's the namespace regex as well

arschles: k, cool

pkg/sshd/config.go
```diff
@@ -1,9 +1,18 @@
...
 type Config struct {
-	SSHHostIP                  string `envconfig:"SSH_HOST_IP" default:"0.0.0.0" required:"true"`
-	SSHHostPort                int    `envconfig:"SSH_HOST_PORT" default:"2223" required:"true"`
-	HealthSrvPort              int    `envconfig:"HEALTH_SERVER_PORT" default:"8092"`
-	HealthSrvTestStorageRegion string `envconfig:"HELATH_SERVER_TEST_STORAGE_REGION" default:"us-east-1"`
+	SSHHostIP                   string `envconfig:"SSH_HOST_IP" default:"0.0.0.0" required:"true"`
+	SSHHostPort                 int    `envconfig:"SSH_HOST_PORT" default:"2223" required:"true"`
+	HealthSrvPort               int    `envconfig:"HEALTH_SERVER_PORT" default:"8092"`
+	HealthSrvTestStorageRegion  string `envconfig:"HEALTH_SERVER_TEST_STORAGE_REGION" default:"us-east-1"`
+	CleanerPollSleepDurationSec int    `envconfig:"CLEANER_POLL_SLEEP_DURATION_SEC" default:"1"`
```
arschles: the only semantic addition here is `CleanerPollSleepDurationSec`, but the spacing of the types changed

bacongobbler:
Seeing my logs being spewed full of messages:

```
Cleaner found no git directories to delete
Cleaner found no git directories to delete
Cleaner found no git directories to delete
Cleaner found no git directories to delete
Cleaner found no git directories to delete
Cleaner found no git directories to delete
Cleaner found no git directories to delete
Cleaner found no git directories to delete
Cleaner found no git directories to delete
```

Then I realized that debug is turned on. Works for me!

arschles:
Note that I've manually tested this and verified it works on GKE

mboersma:
Tested locally on a kube-solo cluster according to the above instructions. Everything went swimmingly. :+1:

pkg/sshd/server.go
```diff
@@ -225,7 +232,8 @@ func (s *server) answer(channel ssh.Channel, requests <-chan *ssh.Request, sshCo
 				}
 
 				repoName := parts[1]
-				if err := buildingRepos.Lock(repoName, time.Duration(0)); err != nil {
+				s.cleanerRef.Lock()
```
arschles: This line will block the cleaner as well as *all other* `git push`es from proceeding. This preserves the goal of preventing multiple git pushes, but reverses the timeout functionality we introduced in #170

smothiki:
Tested Just now working

pkg/sshd/config.go
gerred: Late in the game since this is all around - is this struct tag abuse? Or am I just being a purist here?

gerred:
Tested on GKE as well. :+1:

bacongobbler:
LGTM x 4

pkg/sshd/config.go
krancour: IMO, this is an appropriate use.  Using struct tags to declaratively define the source of data for a struct seems like a typical use case for struct tags.  Router does this extensively as well, using struct tags to define the relationship between struct fields and a map[string]string, as well as constraints upon those fields.

krancour:
Any thoughts of a similar host-level trashman?  Unused / old Docker images can fill up a disk.  We should look into running something like Sherdock in a daemonset in the future.

arschles:
@krancour not sure what you mean by host-level trashman. Can you file an issue with details? If it's something to clean up docker images I'm all for it, but the Sherdock readme doesn't provide much detail.

pkg/sshd/config.go
arschles: Also note that these are standard envconfig-required tags

PR #187, opened 2016-02-18
feat(storage): change outside storage settings

smothiki:
Directly set outside storage URL . example 
```
 - name: "DEIS_OUTSIDE_STORAGE"
   value: "https://storage.googleapis.com"
```
first preference to S3 endpoint then minio

### DIFF
--- a/pkg/gitreceive/storage/endpoint.go
+++ b/pkg/gitreceive/storage/endpoint.go
@@ -6,31 +6,28 @@ import (
 )
 
 const (
-	minioHostEnvVar          = "DEIS_MINIO_SERVICE_HOST"
-	minioPortEnvVar          = "DEIS_MINIO_SERVICE_PORT"
-	outsideStorageHostEnvVar = "DEIS_OUTSIDE_STORAGE_HOST"
-	outsideStoragePortEnvVar = "DEIS_OUTSIDE_STORAGE_PORT"
+	minioHostEnvVar        = "DEIS_MINIO_SERVICE_HOST"
+	minioPortEnvVar        = "DEIS_MINIO_SERVICE_PORT"
+	outsideStorageEndpoint = "DEIS_OUTSIDE_STORAGE"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


arschles:
@smothiki why not leave as-is and if `DEIS_OUTSIDE_STORAGE_HOST` is missing, just ignore it? I like the port option for the customizability.

smothiki:
this PR will give customer to give entire S3 endpoint URL . No need to change two env variables 
endpoint url comes with <protocol>://<host>:<port> . We dont know whether protocol is http or https.
Also if there is custom dns name then port is null

arschles:
Ah, right - missed that. I supposed we can handle custom ports later if needed. marking LGTM1 pending CI

smothiki:
This also defaults to S3EP, so if it is set then need not bother about uninstalling or modifying deis charts

arschles:
Ok. Keep in mind that we'll have to document that fact and add support to the helm charts (via `helm generate`) for operators to set their S3 endpoint (and also their credentials, but that's out of scope of this PR I believe)

smothiki:
Yeah I agree, but I dont think helm has capability to add a new field . Default RC dont have outside storage env at all. And helm supports templates in that case default RC should have outside storage env. Confused with helm here

arschles:
@smothiki helm wouldn't necessarily need to add a new field. If the RC had the following:

```yaml
...
env:
  - name: DEIS_OUTSIDE_STORAGE
    value: {{outsideStorageVal | default ""}}
...
```

Then your code would correctly read the outside storage value if one was present in the `helm generate` command, and otherwise fall back to minio. Unless I read it wrong?

smothiki:
Yup my bad missed it . I'll create an issue to change builder-rc  chart

arschles:
@smothiki :100:

mboersma:
This LGTM once there is an issue or PR to change the builder-rc manifest in the chart.

smothiki:
@mboersma https://github.com/deis/charts/issues/99

PR #188, opened 2016-02-18
chore(manifests): remove repository manifests (Proposal)

arschles:
In the same vein as deis/postgres#38, prefer to use the helm chart manifest to the repository manifests

Ref deis/minio#73


krancour:
Shouldn't there be some complementary changes to the Makefile?  There are targets that use these manifests and I think they're dead weight now.

PR #191, opened 2016-02-18
ref(pkg): move much of the kubernetes code into its own package (WIP)

arschles:
The goal here is to make the k8s code more testable

TODO:

- [ ] Implement tests in `./pkg/k8s/pod_test.go`


arschles:
Closing this as this refactor has become extremely out of date

PR #196, opened 2016-02-23
feat(storage): configure bucket name

smothiki:
Make root bucket configurable. defaults to /git


smothiki:
Adds storage region to dockerfile pod

smothiki:
refer https://github.com/deis/dockerbuilder/issues/24

Issue #199, opened 2016-02-24
Missing builder pod launch


slack:
Image: `smothiki/builder:v2.1`

Builder is hung on git push (and will wait for timeout):
```
k:beef example-go [master]$ g push deis master
Warning: the RSA host key for '[deis.beef.slack.io]:2222' differs from the key for the IP address '[54.200.165.125]:2222'
Offending key for IP in /Users/jhansen/.ssh/known_hosts:109
Matching host key in /Users/jhansen/.ssh/known_hosts:113
Counting objects: 529, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (524/524), done.
Writing objects: 100% (529/529), 54.51 KiB | 0 bytes/s, done.
Total 529 (delta 482), reused 0 (delta 0)
remote: --->
```

It looks like the slugbuilder launched and completed though:
```
deis        22s         22s        1         slugbuild-earthy-instinct-d4c807b8-a94f52b6   Pod                                           Scheduled   {scheduler }                                        Successfully assigned slugbuild-earthy-instinct-d4c807b8-a94f52b6 to ip-10-0-0-98.us-west-2.compute.internal
deis        22s         22s        1         slugbuild-earthy-instinct-d4c807b8-a94f52b6   Pod       implicitly required container POD   Pulled      {kubelet ip-10-0-0-98.us-west-2.compute.internal}   Container image "gcr.io/google_containers/pause:0.8.0" already present on machine
deis        22s         22s        1         slugbuild-earthy-instinct-d4c807b8-a94f52b6   Pod       implicitly required container POD   Created     {kubelet ip-10-0-0-98.us-west-2.compute.internal}   Created with docker id a626b5939b47
deis        22s         22s        1         slugbuild-earthy-instinct-d4c807b8-a94f52b6   Pod       implicitly required container POD   Started     {kubelet ip-10-0-0-98.us-west-2.compute.internal}   Started with docker id a626b5939b47
deis        20s         20s        1         slugbuild-earthy-instinct-d4c807b8-a94f52b6   Pod       spec.containers{deis-slugbuilder}   Started     {kubelet ip-10-0-0-98.us-west-2.compute.internal}   Started with docker id 710055360b5b
[...]

PR #206, opened 2016-02-27
feat(race): change heritage label for every pod launch

smothiki:
Having same labels for every slugbuilder launch makes GC to garbage collect.
This PR fixes terminating behavior for pod. Also we haven't pin pointed the root cause for this behavior.
Listening for event stream wont help in fixing this issue.

### DIFF
--- a/pkg/gitreceive/k8s_util.go
+++ b/pkg/gitreceive/k8s_util.go
@@ -96,8 +96,7 @@ func buildPod(debug, withAuth bool, name, namespace string, env map[string]inter
 			Name:      name,
 			Namespace: namespace,
 			Labels: map[string]string{
-				"heritage": "deis",
-				"version":  "2.0.0-beta",
+				"heritage": name,
 			},
 		},
 	}


smothiki:
fixes https://github.com/deis/builder/issues/199

slack:
Pretty hilarious hack. Really wanted to understand why this works. Did some spelunking.

So Pod GC should trigger when the deleted pod threshold exceeds 12500 pods: https://github.com/kubernetes/kubernetes/blob/v1.1.2/cmd/kube-controller-manager/app/controllermanager.go#L126

Probably explains why we don't see this as much on lightly used, or new clusters. The "fast delete" behavior will most likely show up after you get 12k pods in the stack...

There is a pod sync process that sucks in terminated pods on a loop: https://github.com/kubernetes/kubernetes/blob/v1.1.2/pkg/controller/gc/gc_controller.go#L67-L74

It feels like  pod listing + sorting is impacted by labels in a non-obvious way. And once the deletion starts happening, the GC process gets the pods in a funny order, most recent pods at the top of the list?

Non-overlapping labels might just be moving our pods out of the firing line:

https://github.com/kubernetes/kubernetes/blob/v1.1.2/pkg/controller/gc/gc_controller.go#L89-L91

Haven't yet teased out the interaction between `podStore.List` and `labels.Everything()`...

slack:
The more I think about this, the more I'm convinced that this probably isn't desired k8s behavior.

slack:
Original conversation upstream:
* First Pod GC implementation: kubernetes/kubernetes#14325
* GC Pod discussion: kubernetes/kubernetes#7660

mboersma:
OMG YOU WEREN'T KIDDING. :boom: I thought that had to be unrelated. Wow.

So could we leave `"heritage": "deis"` and instead add `"gc-hack": name` to get the same behavior?

mboersma:
This also works, and seems more descriptive to me:
```diff
diff --git a/pkg/gitreceive/k8s_util.go b/pkg/gitreceive/k8s_util.go
index bfa863b..1fac569 100644
--- a/pkg/gitreceive/k8s_util.go
+++ b/pkg/gitreceive/k8s_util.go
@@ -96,6 +96,7 @@ func buildPod(debug, withAuth bool, name, namespace string, env map[string]inter
                        Name:      name,
                        Namespace: namespace,
                        Labels: map[string]string{
+                               "gc-hack":  name,
                                "heritage": "deis",
                                "version":  "2.0.0-beta",
                        },
```

arschles:
:+1: awesome you guys got to the bottom of this!

arschles:
I've proposed a more comprehensive solution for the future, which would obviously be more involved than this. See https://github.com/deis/builder/issues/207

PR #208, opened 2016-02-29
fix(config.go): increase builder pod wait duration

### LINKED ISSUE #205: BuilderPodWaitDuration should be more than 5 minutes
After the [BuilderPodWaitDuration](https://github.com/deis/builder/blob/f51d3a2fb374f12f04c45d941362c4f5c6c5ae23/pkg/gitreceive/config.go#L30) time is up, the build is considered a failure and the git push fails. Many buildpack deployments (such as ruby with a database migration task or an assets:precompile task) take more than 5 minutes. Even on slow internet speeds a Dockerfile build could more than easily take longer than 5 minutes.

I don't think there should be a wait duration at all for builds.

arschles:
Set it to 15 minutes, which is the same as the Heroku default. See
https://devcenter.heroku.com/articles/slug-compiler

Fixes #205


jchauncey:
@arschles this is ready

arschles:
Thanks @bacongobbler & @jchauncey

Issue #192, opened 2016-02-20
Make the cleaner do finer-grained locking


arschles:
After #186, the cleaner will lock "the world" each time it needs to do a deletion. This behavior can slow down `git push`es while deletions are ongoing. This issue is for making the cleaner lock only the folder (application name) that it's currently deleting, to make `git push` performance better while the cleaner is deleting.

Note that the common case performance after #186 will not be any worse than beforehand. Performance will only be impacted when apps are deleted (and then only a small amount, unless there are a _lot_ of deleted apps)

arschles:
This is no longer necessary because, after #209, the cleaner is lock-free. If/when it begins using locks, this should be re-opened.

PR #209, opened 2016-02-29
feat(lock): change cleaner code to listen to k8s namespsce events

smothiki:
removes locking completely for cleaner. Deletes an App folder  asap when there is a namespace deletion


pkg/sshd/server.go
```diff
@@ -258,7 +258,7 @@ func (s *server) answer(channel ssh.Channel, requests <-chan *ssh.Request, sshCo
 					// TODO: this is an important error case that needs to be covered
 					// Probably the best solution is to change the lock into a lease so that even on unlock failures, RepositoryLock will eventually yield
 				}
-				s.cleanerRef.Unlock()
+				s.cleanerRef.RUnlock()
```
arschles: I believe you also need to change the [`Unlock` call above](https://github.com/deis/builder/pull/209/files#diff-b60cdeda68b99d71a51ec60fc7298c49R244) to `RUnlock`

arschles: see https://github.com/deis/builder/pull/209/files#diff-b60cdeda68b99d71a51ec60fc7298c49R244

smothiki: thanks will make the change ASAP

pkg/cleaner/cleaner.go
```diff
@@ -21,11 +21,19 @@ const (
 )
 
 type Ref struct {
-	mut *sync.Mutex
+	mut *sync.RWMutex
```
technosophos: I know this is something that should have been asked in the earlier commit, but is there a reason we're not just doing `type Ref struct { sync.RWMutex }` and then skipping the four pass-thru methods?

arschles: @technosophos good question. I created the pass-through funcs so that, in theory, https://github.com/deis/builder/issues/192 could be done. It would involve a `map[string]*sync.RWMutex` (or similar) internally.

arschles:
@smothiki tests now fail:

```console
ENG000656:builder aaronschlesinger$ make test
docker run --rm -e GO15VENDOREXPERIMENT=1 -v /Users/aaronschlesinger/gocode/src/github.com/deis/builder:/go/src/github.com/deis/builder -w /go/src/github.com/deis/builder quay.io/deis/go-dev:0.8.0 sh -c 'go test $(glide nv)'
?   	github.com/deis/builder/pkg	[no test files]
ok  	github.com/deis/builder/pkg/cleaner	0.016s
?   	github.com/deis/builder/pkg/conf	[no test files]
ok  	github.com/deis/builder/pkg/controller	0.003s
ok  	github.com/deis/builder/pkg/env	0.002s
?   	github.com/deis/builder/pkg/git	[no test files]
ok  	github.com/deis/builder/pkg/gitreceive	0.005s
ok  	github.com/deis/builder/pkg/gitreceive/git	0.002s
ok  	github.com/deis/builder/pkg/gitreceive/storage	0.004s
ok  	github.com/deis/builder/pkg/healthsrv	0.006s
?   	github.com/deis/builder/pkg/k8s	[no test files]
[info] 2016/02/29 23:00:03 Listening on 127.0.0.1:2244
[info] 2016/02/29 23:00:03 Accepting new connections.
[info] 2016/02/29 23:00:03 Checking closer.
[info] 2016/02/29 23:00:03 Checking closer.
[info] 2016/02/29 23:00:03 Accepted connection.
[info] 2016/02/29 23:00:03 Channel type: session
Key='HELLO', Value='world'
[info] 2016/02/29 23:00:03 PING
[info] 2016/02/29 23:00:03 Channel type: session
[warning] 2016/02/29 23:00:03 Illegal command is 'illegal'
[info] 2016/02/29 23:00:03 Listening on 127.0.0.1:2245
[info] 2016/02/29 23:00:03 Accepting new connections.
[info] 2016/02/29 23:00:03 Checking closer.
[info] 2016/02/29 23:00:03 Checking closer.
[info] 2016/02/29 23:00:03 Accepted connection.
[info] 2016/02/29 23:00:04 Channel type: session
[warning] 2016/02/29 23:00:04 Expected two-part command.
[info] 2016/02/29 23:00:04 Channel type: session
[info] 2016/02/29 23:00:06 Channel type: session
[error] 2016/02/29 23:00:06 Another git push is ongoing
[info] 2016/02/29 23:00:06 Listening on 127.0.0.1:2246
[info] 2016/02/29 23:00:06 Accepting new connections.
[info] 2016/02/29 23:00:06 Checking closer.
[info] 2016/02/29 23:00:06 Checking closer.
[info] 2016/02/29 23:00:06 Accepted connection.
[info] 2016/02/29 23:00:06 Channel type: session
[warning] 2016/02/29 23:00:06 Expected two-part command.
[info] 2016/02/29 23:00:06 Channel type: session
[error] 2016/02/29 23:00:06 Another git push is ongoing
--- FAIL: TestDelete (0.21s)
	server_test.go:207: expected no error, got ssh: command git-upload-pack /demo.git failed
FAIL
FAIL	github.com/deis/builder/pkg/sshd	3.638s
?   	github.com/deis/builder	[no test files]
make: *** [test] Error 1
```

pkg/cleaner/cleaner.go
```diff
@@ -2,131 +2,33 @@
...
-			log.Printf("Cleaner error listing namespaces (%s)", err)
-			continue
-		} else {
-			lst := make([]string, len(nsList.Items))
-			for i, ns := range nsList.Items {
-				lst[i] = strings.ToLower(ns.Name)
+		event := <-watcher.ResultChan()
+		namespace := event.Object.(*api.Namespace)
+		if event.Type == "DELETED" {
+			if err := os.RemoveAll(namespace.ObjectMeta.Name + dotGitSuffix); err != nil {
```
arschles: If a folder is deleted while a `git push` operation is in progress, the behavior of the builder is essentially undefined and the user will see one of quite a few error messages, none of which make sense. I saw this behavior in effect when I first built the cleaner, before there was locking.

That needs to be fixed in some way. The most obvious way is to use a lock, but since the description here says `removes locking completely for cleaner`, I'm guessing you don't want to go there ;). Instead, perhaps you could use a `WaitGroup` that the cleaner increments when it's about to do its work, decrements when it's done, and the SSH server waits for?

smothiki: Builder state becomes inconsistent for a deleted App's git push failure is fine cause any ways the git push will fail while doing config, build or push hook. The only solution to make it cleaner is to have a hook from controller to check if the APP exists and then proceed accordingly. This can be handled later.

arschles: >Builder state becomes inconsistent for a deleted App's git push failure is fine cause any ways

I disagree. The inconsistencies spread beyond builder, for example. Say a directory is deleted during a slugbuilder's run. After the run is done, the tarball and slug will be in object storage, and the build will have been started in workflow, but it will not be finished. Also, from a user perspective, the error message will indicate a failed call to the controller - that is, it'll make no sense.

>The only solution to make it cleaner is to have a hook from controller to check if the APP exists and then proceed accordingly

The builder would have to make this call before every state-changing operation, and even then it still could result in inconsistent states.

>This can be handled later

It can, but it shouldn't in my opinion, for the above reasons.

pkg/cleaner/cleaner.go
```diff
@@ -2,131 +2,34 @@
...
-		} else {
-			lst := make([]string, len(nsList.Items))
-			for i, ns := range nsList.Items {
-				lst[i] = strings.ToLower(ns.Name)
+		event := <-watcher.ResultChan()
+		namespace := event.Object.(*api.Namespace)
+		appToDelete := gitHome + "/" + namespace.ObjectMeta.Name + dotGitSuffix
+		if event.Type == "DELETED" {
+			if err := os.RemoveAll(appToDelete); err != nil {
+				log.Printf("Cleaner error removing deleted app %s (%s)", appToDelete, err)
```
arschles: @smothiki can you write some tests for this logic?

smothiki: Sure

arschles:
@smothiki just tested. A concurrent `git push` to the same app results in this:

```console
ENG000656:example-go aaronschlesinger$ git push deis master
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Otherwise, deleting an app, re-creating it, and pushing a different repo to the same app name seems to work.

arschles:
Also, CI will say this in a few, but tests don't compile anymore:

```console
ENG000656:builder aaronschlesinger$ make test
docker run --rm -e GO15VENDOREXPERIMENT=1 -v /Users/aaronschlesinger/gocode/src/github.com/deis/builder:/go/src/github.com/deis/builder -w /go/src/github.com/deis/builder quay.io/deis/go-dev:0.8.0 sh -c 'go test $(glide nv)'
?   	github.com/deis/builder/pkg	[no test files]
?   	github.com/deis/builder/pkg/cleaner	[no test files]
?   	github.com/deis/builder/pkg/conf	[no test files]
ok  	github.com/deis/builder/pkg/controller	0.003s
ok  	github.com/deis/builder/pkg/env	0.002s
?   	github.com/deis/builder/pkg/git	[no test files]
ok  	github.com/deis/builder/pkg/gitreceive	0.007s
ok  	github.com/deis/builder/pkg/gitreceive/git	0.002s
ok  	github.com/deis/builder/pkg/gitreceive/storage	0.003s
ok  	github.com/deis/builder/pkg/healthsrv	0.004s
?   	github.com/deis/builder/pkg/k8s	[no test files]
# github.com/deis/builder/pkg/sshd
pkg/sshd/server_test.go:273: undefined: cleaner.NewRef
pkg/sshd/server_test.go:275: too many arguments in call to Serve
FAIL	github.com/deis/builder/pkg/sshd [build failed]
?   	github.com/deis/builder	[no test files]
make: *** [test] Error 2
```

jchauncey:
I checked out this branch and did a concurrent push and everything seemed to be ok

jchauncey:
nevermind i forgot to build a fresh binary.

smothiki:
this issue is already present in the master branch . As we have avoided global lock for each exec req we haven;t seen this till now 
https://github.com/deis/builder/issues/217

arschles:
@smothiki ok, thanks. that being said, this PR looks ok to merge. I'm LGTMing

arschles:
@smothiki I forgot about my previous comment about writing tests for the cleaner, sorry. removing my LGTM for now. please let me know when tests are done and I'll re-review

smothiki:
refer https://github.com/deis/builder/issues/228

arschles:
@smothiki does this need another manual test since the last one?

smothiki:
This is working no need of testing Just tested the behavior

pkg/k8s/namespace.go
```diff
@@ -14,4 +15,5 @@ import (
 //	nsl = kubeClient.Namespaces()
 type NamespaceLister interface {
 	List(labels.Selector, fields.Selector) (*api.NamespaceList, error)
+	Watch(label labels.Selector, field fields.Selector, resourceVersion string) (watch.Interface, error)
```
arschles: @smothiki this should probably go into a separate interface called `NamespaceWatcher` or similar

smothiki: sure

mboersma:
Code looks good on first pass, testing it in k8s now.

mboersma:
This works great. I had three simultaneous pushes going from a combination of Dockerfile and Procfile apps, and nothing ever blocked or errored. A shell inside the builder container shows everything was cleaned up immediately under /home/git. I kept creating, pushing, deleting simultaneously but wasn't able to cause a failure.

Nice work!

PR #211, opened 2016-02-29
fix(pkg/sshd/server_test.go): write test for gitPktLine (WIP)

arschles:


pkg/sshd/server_test.go
```diff
@@ -8,6 +10,7 @@ import (
 	"github.com/deis/builder/pkg/controller"
 
 	"github.com/Masterminds/cookoo"
+	"github.com/arschles/assert"
```
jchauncey: dont want to be a stickler here but should this be in pkg or can we use antoher matching lib like gomega?

arschles: >dont want to be a stickler here but should this be in pkg

I don't think so - we could fork it to the Deis org if you like. I've used this elsewhere as-is though.

>can we use antoher matching lib like gomega?

I'd prefer not to in this particular case because Gomega is a larger dependency that includes a lot of functionality not needed here.

jchauncey: k my only concern is the day you leave =P and we are dependent on a package we dont control

arschles: Ok, cool with you if I put in an issue to fork it & change the deps? It's a fairly far-reaching change in this repo.

Also, fwiw, we're dependent on lots of packages we don't control ;)

jchauncey: yeah thats fine. =)

arschles: https://github.com/deis/builder/issues/216

PR #223, opened 2016-03-02
fix(pkg/gitreceive): replace the AWS SDK with minio-go

### LINKED ISSUE #222: aws s3 go sdk only works for minio and aws S3
the cli code to access S3 works only for AWS S3 and minio .
Tested with goamz/goamz S3 client which works for ceph , aws, gcs but not minio .
Tested with AdRoll/goamz which is slight modification of above mentioned repo which works for Aws S3, GCS, minio but not for ceph.

For beta release atleast GCS should work . So we have to implement AdRoll goamz

arschles:
To support Google Cloud Storage compatibility mode, the builder must support AWS version 2 signatures. Singe the `github.com/aws/aws-sdk-go` package package doesn't support v2 signing for S3, this PR replaces that package with `github.com/minio/minio-go`, which does.

This PR also emphasizes testing related to object storage, and while necessary, that emphasis led to more changes here than strictly necessary to achieve the functionality we needed.

See below for a list of additional changes necessary to aid testing:

- `minio-go` compatible interfaces
- Mocks for each of the above interfaces
- Unit tests for most of the `func`s inside of `github.com/deis/builder/pkg/gitreceive/storage`
- More logging in the git receive hook
- New utility functions (and tests for them) on `SlugBuilderInfo`
- A new `createBuildHook` in `pkg/gitreceive/controller.go`, and tests for it

Fixes #222 

TODO: After this is done, create an issue to:

- ~~Move the object storage logic into one package: `github.com/deis/builder/pkg/storage`~~ - it's been created at https://github.com/deis/builder/issues/226
- ~~Refactor the health check endpoint code to use minio-go~~ - that had to be done in this PR

### DIFF
--- a/pkg/gitreceive/storage/client.go
+++ b/pkg/gitreceive/storage/client.go
@@ -1,14 +1,12 @@
 package storage
 
 import (
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/session"
-	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/deis/builder/pkg/sys"
+	s3 "github.com/minio/minio-go"
 )
 
 // GetClient returns a S3 API compatible storage client
-func GetClient(regionStr string, fs sys.FS, env sys.Env) (*s3.S3, error) {
+func GetClient(regionStr string, fs sys.FS, env sys.Env) (*s3.Client, error) {
 	auth, err := getAuth(fs)
 	if err != nil {
 		return nil, err
@@ -19,10 +17,5 @@ func GetClient(regionStr string, fs sys.FS, env sys.Env) (*s3.S3, error) {
 		return nil, err
 	}
 
-	return s3.New(session.New(&aws.Config{
-		Credentials:      auth,
-		Region:           aws.String(regionStr),
-		Endpoint:         aws.String(endpoint),
-		S3ForcePathStyle: aws.Bool(true),
-	})), nil
+	return s3.New(endpoint, auth.accessKeyID, auth.accessKeySecret, false), nil
 }
--- a/pkg/gitreceive/storage/client.go
+++ b/pkg/gitreceive/storage/client.go
@@ -11,13 +11,8 @@ type Client struct {
 }
 
 // GetClient returns a S3 API compatible storage client
-<<<<<<< 7c8dc9bdf6ffed171ed952661229ae2b8590cef7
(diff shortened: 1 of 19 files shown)


pkg/gitreceive/storage/client.go
```diff
@@ -18,10 +21,10 @@ func GetClient(regionStr string) (*s3.S3, error) {
...
 
-	return s3.New(session.New(&aws.Config{
-		Credentials:      auth,
-		Region:           aws.String(regionStr),
-		Endpoint:         aws.String(endpoint),
-		S3ForcePathStyle: aws.Bool(true),
-	})), nil
+	// the New function call guesses which signature version to use. Currently, it correctly guesses V2 for GCS and V4 for both AWS S3 and Minio
+	s3Client, err := s3.New(endpoint.URLStr, auth.accessKeyID, auth.accessKeySecret, !endpoint.Secure)
+	if err != nil {
```
smothiki: client signature should change according to the storage back end we are accessing . For minio it should be s3.NewV4. for other s3.NewV2

arschles: note the comment above the `New` function call. the minio-go library takes care of inferring it for us

arschles:
My most recent manual test (done with the [example-go repo](https://github.com/deis/example-go)) correctly launches a slugbuilder, and the slugbuilder correctly downloads the tarball and builds it. When it tries to upload the slug, it hangs however. This is the log line on which it hangs:

`+ mc --quiet -C /app/mc_config cp /tmp/slug.tgz http://10.171.248.9:9000/git/home/gotest:git-49f11b24/push/`

pkg/gitreceive/storage/client.go
smothiki: oops . my bad

smothiki: good to know might be helpful for testing Ceph

smothiki:
Will do manual testing but so far code looks good.

smothiki:
worked with GCS

arschles:
Can you try a slug build with minio as well? My last test showed it hanging and I'd like to see if that's systemic

smothiki:
removed my LGTM as checking for slug files is timing out.

arschles:
Status as of now: `git push`es using minio as the backing object store now succeed, and the resultant app (running in a slugrunner) responds to requests properly.

smothiki:
Just tested working

arschles:
This is tested (again) as working post-rebase

PR #229, opened 2016-03-03
fix(*): creating and using interfaces for system tasks

### LINKED ISSUE #227: Increase test coverage in ./pkg/gitreceive/storage
Several important `func`s are untested because they interact with the filesystem or environment. Here's a non-exhaustive list:

- `GetClient`
- `getEndpoint`
- `getAuth`
- `CredsOK` (relies on `getAuth`)

In order to test them, put the I/O behind interfaces that can be mocked.

If #226 is done before this, the storage package will be moved to `./pkg/storage`, but this issue remains otherwise the same.

arschles:
This PR creates interfaces meant for interacting with environment variables and for reading from the filesystem. Various pieces of the storage package use these interfaces, and this PR also adds mock implementations of those interfaces and unit tests for the aforementioned pieces.

Fixes #227 

TODO: 

- [x] Create interfaces, real and mock implementations for system tasks
- [x] Add tests for functionality that now uses the aforementioned interfaces

### DIFF
--- a/pkg/gitreceive/build.go
+++ b/pkg/gitreceive/build.go
@@ -15,6 +15,7 @@ import (
 	"github.com/deis/builder/pkg"
 	"github.com/deis/builder/pkg/gitreceive/git"
 	"github.com/deis/builder/pkg/gitreceive/storage"
+	"github.com/deis/builder/pkg/sys"
 	"github.com/deis/pkg/log"
 	"gopkg.in/yaml.v2"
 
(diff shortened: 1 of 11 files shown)


pkg/gitreceive/build.go
```diff
@@ -41,7 +42,7 @@ func run(cmd *exec.Cmd) error {
 	return cmd.Run()
 }
 
-func build(conf *Config, s3Client *s3.S3, kubeClient *client.Client, builderKey, rawGitSha string) error {
+func build(conf *Config, s3Client *s3.S3, kubeClient *client.Client, env sys.Env, fs sys.FS, builderKey, rawGitSha string) error {
```
jackfrancis: This is ticky-tacky, but... all other functions that accept both a sys.FS and sys.ENV interface do so in according to an ordered arity that looks like `(<params>...sys.FS, sys.ENV,...<params>)`. Maybe switch the order of `env sys.Env` and `fs sys.FS` params in `build()` above.

arschles: @jackfrancis fair point, change incoming

jackfrancis:
Looks great, interfaces ftw.

smothiki:
More interfaces awsome

Issue #238, opened 2016-03-07
Investigate TestWaitForObjectMissing timing issue


arschles:
see https://travis-ci.org/deis/builder/builds/113820760:

```console
--- FAIL: TestWaitForObjectMissing (0.01s)
	assert.go:40: /go/src/github.com/deis/builder/pkg/gitreceive/storage/object_test.go:92: the statter wasn't called at least once
```

arschles:
Reopening, as this was accidentally closed by https://github.com/deis/builder/pull/241. This issue is not fixed.

PR #241, opened 2016-03-08
fix(pkg/gitreceive/storage/object_test.go): skip TestWaitForObjectMissing

arschles:
only skip it for now. this issue does *not* fix https://github.com/deis/builder/issues/238

cc/ @smothiki


arschles:
@smothiki @kmala I'm going to be waiting for unit tests to finish here before merging

PR #242, opened 2016-03-09
feat(cleaner): write unit test for cleaner 

smothiki:
Little tricky as my first run got hung up at watching events at the channel
Working more on this.

### DIFF
--- a/pkg/cleaner/cleaner.go
+++ b/pkg/cleaner/cleaner.go
@@ -3,20 +3,21 @@ package cleaner
 
 import (
 	"log"
-	"os"
 
 	"k8s.io/kubernetes/pkg/api"
 
 	"github.com/deis/builder/pkg/k8s"
+	"github.com/deis/builder/pkg/sys"
 )
 
 const (
 	dotGitSuffix = ".git"
 )
 
-// Run starts the deleted app cleaner. Every pollSleepDuration, it compares the result of nsLister.List with the directories in the top level of gitHome on the local file system. On any error, it uses log messages to output a human readable description of what happened.
-func Run(gitHome string, nsLister k8s.NamespaceWatcher) error {
-
+// Run starts the deleted app cleaner. This function listens to the Kubernetes event stream for events that indicate namespaces that were `DELETED`.
+// If the namespace name matches a folder on the local filesystem, this func deletes that folder.
+// Note that this function blocks until the watcher returned by `nsLister.Watch` is closed, so you should launch it in a goroutine.
+func Run(gitHome string, nsLister k8s.NamespaceWatcher, fs sys.FS) error {
 	watcher, err := nsLister.Watch(nil, nil, "")
 	if err != nil {
 		log.Printf("unable to get watch events (%s)", err)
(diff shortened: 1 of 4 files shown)


pkg/cleaner/cleaner.go
```diff
@@ -3,20 +3,19 @@ package cleaner
...
 )
 
 const (
 	dotGitSuffix = ".git"
 )
 
-// Run starts the deleted app cleaner. Every pollSleepDuration, it compares the result of nsLister.List with the directories in the top level of gitHome on the local file system. On any error, it uses log messages to output a human readable description of what happened.
-func Run(gitHome string, nsLister k8s.NamespaceWatcher) error {
-
+// Run starts the deleted app cleaner. Every pollSleepD uration, it compares the result of nsLister.List with the directories in the top level of gitHome on the local file system. On any error, it uses log messages to output a human readable description of what happened.
```
arschles: @smothiki these docs are out of date now. Can you fix them to reflect the fact that it listens to the event stream. Maybe something like this:

```
Run starts the deleted app cleaner. This function listens to the Kubernetes event stream for events that indicate namespaces that were `DELETED`. If the namespace name matches a folder on the local filesystem, this func deletes that folder. Note that this function blocks until the watcher returned by `nsLister.Watch` is closed, so you should launch it in a goroutine.
```

pkg/cleaner/cleaner.go
```diff
@@ -26,7 +25,7 @@ func Run(gitHome string, nsLister k8s.NamespaceWatcher) error {
 		if event.Type == "DELETED" {
 			namespace := event.Object.(*api.Namespace)
```
arschles: @smothiki you should check this type assertion, and if it wasn't an `*api.Namespace`, then don't continue.

pkg/cleaner/cleaner_test.go
```diff
@@ -0,0 +1,44 @@
...
+	fs := sys.NewFakeFS()
+	dirhome := "/home/git"
+	fs.Files["/home/git/dir1.git"] = []byte{}
+	go Run(dirhome, ns, fs)
+	time.Sleep(5 * time.Second)
+	_, ok := fs.Files["/home/git/dir1.git"]
+	if ok {
+		t.Fatal("direcotry not deleted unit test failed")
+	}
+}
```
arschles: @smothiki can you add a few other tests for failure cases. For example:

1. The watcher sends other objects, like pods or replication controllers. In those cases, the cleaner should do nothing
2. The watcher sends namespaces that don't exist in the fake filesystem
3. The watcher doesn't send any `DELETE` events

pkg/cleaner/cleaner_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"k8s.io/kubernetes/pkg/fields"
+	"k8s.io/kubernetes/pkg/labels"
+	"k8s.io/kubernetes/pkg/watch"
+)
+
+type namespace struct {
+}
+
+// Watch returns a watch.Interface that watches the requested namespaces.
+
```
arschles: nitpick: can you remove this line break

pkg/cleaner/cleaner_test.go
```diff
@@ -0,0 +1,44 @@
...
+	go func() {
+		nst.Add(&api.Namespace{ObjectMeta: api.ObjectMeta{Name: "dir1"}})
+		nst.Modify(&api.Namespace{ObjectMeta: api.ObjectMeta{Name: "dir1"}})
+		nst.Delete(&api.Namespace{ObjectMeta: api.ObjectMeta{Name: "dir1"}})
+		nst.Stop()
+	}()
+	return nst, nil
+}
+
+func (r *namespace) IsAnAPIObject() {}
```
arschles: doesn't this need to return `bool`?

smothiki:
@arschles thanks for the comments . I'm updating the PR accordingly

pkg/cleaner/cleaner_test.go
smothiki: IsANAPIObject doesn;t return anything. It just has a pointer receiver

arschles:
@smothiki thanks - let me know when you'd like another review

smothiki:
@arschles  let me know if we can delete testdata dir

arschles:
@smothiki I reviewed again, and only [this comment](https://github.com/deis/builder/pull/242#discussion_r55739907) remains. Also, is the testdata directory used anywhere else?

PR #243, opened 2016-03-09
feat(builder): get the slug/docker builder images from the environment

### LINKED ISSUE #66: Make slugbuilder and dockerimage images configurable
Adapted from deis/workflow#135

The image used for building slugs should be configurable via an environment variable (e.g. `SLUGBUILDER_IMAGE_NAME`) so that a cluster administrator can specify their own image to use. If the environment variable is missing, builder should default to the `quay.io/deisci/slugbuilder` image.

The same requirement will likely apply to the dockerbuilder image, but that's out of scope of this issue.

If so, we have to consider if someone specifies a slugbuilder or dockerbuilder image that's on the internal registry. For example, what happens when someone specifies an image that lives on the internal docker registry, the `docker pull` is in progress, and the registry crashes?

kmala:
fixes #66 

*Update by mboersma*:
Closes deis/workflow#135.
Closes deis/workflow#323.

### DIFF
--- a/pkg/gitreceive/k8s_util.go
+++ b/pkg/gitreceive/k8s_util.go
@@ -12,10 +12,8 @@ import (
 )
 
 const (
-	slugBuilderName    = "deis-slugbuilder"
-	slugBuilderImage   = "quay.io/deisci/slugbuilder:v2-beta"
-	dockerBuilderName  = "deis-dockerbuilder"
-	dockerBuilderImage = "quay.io/deisci/dockerbuilder:v2-beta"
+	slugBuilderName   = "deis-slugbuilder"
+	dockerBuilderName = "deis-dockerbuilder"
 
 	tarURLKey        = "TAR_URL"
 	putURLKey        = "put_url"
@@ -35,7 +33,7 @@ func slugBuilderPodName(appName, shortSha string) string {
 	return fmt.Sprintf("slugbuild-%s-%s-%s", appName, shortSha, uid)
 }
 
-func dockerBuilderPod(debug, withAuth bool, name, namespace string, env map[string]interface{}, tarURL, imageName, region string) *api.Pod {
+func dockerBuilderPod(debug, withAuth bool, name, namespace string, env map[string]interface{}, tarURL, imageName, region, dockerBuilderImage string) *api.Pod {
 	pod := buildPod(debug, withAuth, name, namespace, env)
 
 	pod.Spec.Containers[0].Name = dockerBuilderName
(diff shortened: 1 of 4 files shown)


pkg/gitreceive/k8s_util.go
```diff
@@ -25,6 +24,11 @@ const (
 	dockerSocketPath = "/var/run/docker.sock"
 )
 
+var (
+	slugBuilderImage   = getenv("SLUGBUILDER_IMAGE_NAME", "quay.io/deisci/slugbuilder:v2-beta")
+	dockerBuilderImage = getenv("DOCKERBUILDER_IMAGE_NAME", "quay.io/deisci/dockerbuilder:v2-beta")
```
arschles: @kmala can you use the [envconfig](https://github.com/kelseyhightower/envconfig) [struct](https://github.com/deis/builder/blob/master/pkg/gitreceive/config.go#L13-L34) that we have already to get these environment variables?

arschles: also, the change I mentioned above will require that the values get passed down into the `dockerBuilderPod` and `slugBuilderPod` `func`s, so can you add to appropriate tests to ensure that the correct images are set?

kmala: @arschles can you check now again?

pkg/gitreceive/k8s_util_test.go
```diff
@@ -53,18 +55,18 @@ func TestBuildPod(t *testing.T) {
...
-		{true, true, "test", "default", env, "tar", "put-url", "buildpack"},
-		{true, false, "test", "default", env, "tar", "put-url", "buildpack"},
+		{true, true, "test", "default", emptyEnv, "tar", "put-url", "", ""},
+		{true, false, "test", "default", emptyEnv, "tar", "put-url", "", ""},
+		{true, true, "test", "default", env, "tar", "put-url", "", ""},
+		{true, false, "test", "default", env, "tar", "put-url", "", ""},
+		{true, true, "test", "default", emptyEnv, "tar", "put-url", "buildpack", ""},
+		{true, false, "test", "default", emptyEnv, "tar", "put-url", "buildpack", ""},
+		{true, true, "test", "default", env, "tar", "put-url", "buildpack", ""},
+		{true, false, "test", "default", env, "tar", "put-url", "buildpack", ""},
```
arschles: could you add a row here to test for a custom image name? same with the below table for docker builds

pkg/gitreceive/k8s_util.go
arschles: @kmala just went over it. only https://github.com/deis/builder/pull/243#discussion_r55615726 remains. thanks for making the changes

pkg/gitreceive/k8s_util_test.go
kmala: added tests

arschles:
@kmala code LGTM. I haven't run any manual tests though

PR #258, opened 2016-03-15
chore(coverage): add codecov.io integration (WIP)

arschles:


test-cover.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -e
+echo "" > coverage.txt
+
+for d in $(find . -maxdepth 10 -type d); do
```
mboersma: What about:
```bash
for d in $(go list $(glide novendor)); do
```

mboersma:
See also #314.

arschles:
bumping from RC1, as this is not critical for the RC

arschles:
Closing in favor of #314

Issue #224, opened 2016-03-02
Use deis-objectstorage-secret.yaml to get S3 credentials


smothiki:
https://github.com/deis/charts/blob/master/deis-dev/tpl/deis-objectstorage-secret.yaml

the above secret already includes   bucket and region

arschles:
@smothiki it seems that builder, along with the chart, functions properly, and can be configured as-is. My vote is to remove this from the showstoppers list

arschles:
[Object storage docs](https://github.com/deis/workflow/blob/master/src/installing-workflow/configuring-object-storage.md#credentials) should also be upgraded when this work is done.

krancour:
> Object storage docs should also be upgraded when this work is done.

Related: https://github.com/deis/workflow/issues/61

arschles:
Related issues: https://github.com/deis/slugbuilder/issues/32 and https://github.com/deis/slugrunner/issues/18

kmala:
fixed in #262

PR #262, opened 2016-03-16
feat(storage): implement separate storage layer

kmala:
This PR is to create a separate storage layer for the deis builder so that it can support multiple storage backends.


Makefile
```diff
@@ -25,7 +25,7 @@ DEIS_REGISTRY ?= ${DEV_REGISTRY}/
 # Kubernetes-specific information for RC, Service, and Image.
 RC := manifests/deis-${SHORT_NAME}-rc.yaml
 SVC := manifests/deis-${SHORT_NAME}-service.yaml
-IMAGE := ${DEIS_REGISTRY}${IMAGE_PREFIX}/${SHORT_NAME}:${VERSION}
+IMAGE := ${DEIS_REGISTRY}/${SHORT_NAME}:${VERSION}
```
arschles: why remove the `IMAGE_PREFIX`?

kmala: That is just for my local testing.I need to rebase this and clean a lot of changes before it becomes mergable

arschles:
@kmala since slugrunner, slugbuilder and dockerbuilder are gonna need this storage layer, wouldn't it be better to put it into a separate repository and also build a CLI for it (for the shell components to use)?

krancour:
@arschles does some such abstraction already exist?  It seems to me there are several.  s3cmd comes to mind as one example.  Do they all fall short?

bacongobbler:
Azure is not supported by s3cmd, for example. I'd imagine this being like a generic key/value store library for the three components.

kmala:
@krancour I think we should we use the native client libraries instead of using some common abstraction because first they may not support everything and second it will be hard to support future requirements when we might need to extend to some other storage backends.

arschles:
:+1: to what @bacongobbler said. I'd love to be able to interact with azure, gcs, s3 and others all from the same CLI. See below for a rough outline of what this mythical CLI might be able to do.

```console
# interact with Azure
deis-obj-storage register azure https://someurl.com mycreds
deis-obj-storage upload azure/mybucket/myobj
# interact with GCS
deis-obj-storage register standard_gcs https://storage.googleapis.com mycreds
deis-obj-storage upload standard_gcs/mybucket/myobj
# interact with S3
deis-obj-storage register standard_s3 https://s3.amazonaws.com
deis-obj-storage upload standard_s3/mybucket/myobj
# also, support more obj storage systems in the future?
```

kmala:
fixes #224

versioning.mk
```diff
@@ -1,6 +1,6 @@
 MUTABLE_VERSION ?= canary
 VERSION ?= git-$(shell git rev-parse --short HEAD)
-IMAGE_PREFIX ?= deis
+IMAGE_PREFIX ?=
```
arschles: @kmala this needs a value

rootfs/Dockerfile
```diff
@@ -3,8 +3,7 @@ FROM alpine:3.3
 # specify the git user's home and name
 ENV GITHOME /home/git
 ENV GITUSER git
-# this is so the minio client (https://github.com/minio/mc) works properly
-ENV DOCKERIMAGE=1
```
arschles: :+1:

kmala:
fixes #266

glide.yaml
```diff
@@ -32,4 +32,6 @@ import:
   version: ~1.1
 - package: github.com/arschles/assert
   version: 6882f85ccdc7c1822b146d1a6b0c2c48f91b5140
-- package: github.com/minio/minio-go
+- package: github.com/docker/distribution
+  repo: https://github.com/kmala/distribution
```
krancour: There's a Deis fork of this... you should probably get whatever customizations you've made merged into there, then this should reference that rather than your own fork.

kmala: yes...i already have a PR for that https://github.com/deis/distribution/pull/7
Once it is merged i will change

boot.go
```diff
@@ -11,11 +11,15 @@ import (
...
-	"github.com/deis/builder/pkg/gitreceive/storage"
 	"github.com/deis/builder/pkg/healthsrv"
 	"github.com/deis/builder/pkg/sshd"
 	"github.com/deis/builder/pkg/sys"
 	pkglog "github.com/deis/pkg/log"
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
+	_ "github.com/docker/distribution/registry/storage/driver/azure"
+	"github.com/docker/distribution/registry/storage/driver/factory"
+	_ "github.com/docker/distribution/registry/storage/driver/gcs"
+	_ "github.com/docker/distribution/registry/storage/driver/s3-aws"
```
krancour: I get what you're going for here... most of what you need is already implemented in these packages... but it feels very "weird" to rely on these packages that have a great deal to do with registry when the builder code really has nothing to do with registry.  Can the useful bits of Docker's code be split into a separate project that can be a standalone go-based object storage library?

arschles: +1 to @krancour's statement. We should at least use [our fork](https://github.com/deis/distribution), but I'd prefer a new repository that is 100% dedicated to this storage abstraction. Note that @mboersma has submitted https://github.com/docker/distribution/issues/1537, which is highly related.

Finally, I'd prefer that, eventually, we refactor whatever storage library that we work with to:

- Not use anonymous imports unless for good, documented reason (I don't see any documentation anywhere at or under https://godoc.org/github.com/docker/distribution/registry/storage/driver)
- Not use the factory pattern (I don't see a use for it based on https://godoc.org/github.com/docker/distribution/registry/storage/driver/factory)

kmala: I am using  [our fork](https://github.com/deis/distribution) only. I had aliased docker distribution to grab our repo in the glide.yaml file the reason being i don't want to change all the imports in our repo.
The plan is to initially use it this way and depending on the PR in the docker distribution we want to change ourself making into a separate repo.

arschles: @kmala responses below:

>I am using our fork only. I had aliased docker distribution to grab our repo in the glide.yaml file the reason being i don't want to change all the imports in our repo.

Ah, got it. Sorry I didn't realize what's going on there.

>The plan is to initially use it this way and depending on the PR in the docker distribution we want to change ourself making into a separate repo.

That plan is fine with me for now. FWIW, my guess is that docker/distribution#1537 is going to be a "won't fix".

stevvooe: > That plan is fine with me for now. FWIW, my guess is that docker/distribution#1537 is going to be a "won't fix".

Don't count us out. We'd like to avoid extra maintenance overhead of a separate project. However, if people find the model to be effective and some extra work is done to increase fitness for external usage, we will likely reconsider. There really is nothing registry specific about these packages.

> Not use anonymous imports unless for good, documented reason

These are to support back registration. If you find the documentation to be lacking, we are willing to accept a PR. Up until this point, the usage of this package has been mostly internal. If we split out this project, the back registration would likely remain in the registry project, as it is very specific to the configuration environment. There is no reason to continue to maintain it in this manner, other than convenience, so we're open to suggestions.

arschles: @stevvooe I certainly haven't counted you out (I was simply speculating), and definitely understand the maintenance overhead of the separate project.

Reading up, I realize that I wasn't clear that we can make progress with, and will submit PRs upstream to either deis/distribution or the new repo if you do decide to split it out. This storage abstraction is useful to us either way, and a new repo would just be a convenience (precisely because the storage layer is not registry specific).

>These are to support back registration. If you find the documentation to be lacking, we are willing to accept a PR.

Great - I'll need to read through the registry codebase more thoroughly to know exactly what you mean by back registration. Mind pointing me to a good starting place?

>There is no reason to continue to maintain it in this manner, other than convenience, so we're open to suggestions.

Just so I'm clear, do you mean there would be no reason to continue to maintain it in this manner if you do split out the project, or is there something I'm missing?

Anyway, I'm happy to help out where possible (beyond the docs mentioned above).

stevvooe: A solid place to start with back registration is [here](https://github.com/docker/distribution/blob/master/registry/storage/driver/s3-aws/s3.go#L88). The factory is registered such that the driver can be instantiated by name and opaque configuration values. The anonymous imports (ie `_`) force the registration via the `init` method and make it available. It allows one to have [different mains](https://github.com/docker/distribution/blob/master/cmd/registry/main.go) to compile in different sets of driver plugins. We used to use this, along with build tags, to control which drivers were compiled into the binary. This is a pattern that is leveraged in several places, such as `crypto` hash functions and `database/sql`.

What I meant by my word soup was that we can separate the back registration from the driver implementations. Registry specific configuration is present in a lot of the driver packages that we might want to have a closer look at. The exact split is unclear, but we would go with what would make it broadly useful.

arschles: >A solid place to start with back registration is here. The factory is registered such that the driver can be instantiated by name and opaque configuration values. The anonymous imports (ie _) force the registration via the init method and make it available

Thanks for that [this documentation PR](https://github.com/docker/distribution/pull/1546) should be a decent (if small) start to documenting this registration mechanism.

>What I meant by my word soup was that we can separate the back registration from the driver implementations. 

That's great!

>Registry specific configuration is present in a lot of the driver packages that we might want to have a closer look at. The exact split is unclear, but we would go with what would make it broadly useful.

I'll have a look to get myself more familiar on the registry bits. My general thought was to move just `StorageDriver` and implementations out, leaving the factory in. FWIW

arschles:
@kmala I just merged https://github.com/deis/builder/pull/248, which is gonna cause you some rebasing pain. Let me know when you hit that and I'll help out

glide.yaml
arschles: @kmala I think what @krancour was referring to is that you use the repo of the Deis fork - https://github.com/deis/distribution

pkg/conf/config.go
```diff
@@ -35,3 +41,35 @@ func GetBuilderKey() (string, error) {
 	builderKey := string(builderKeyBytes)
 	return builderKey, nil
 }
+
+func GetStorageParams(env sys.Env) (Parameters, error) {
```
arschles: I'd like this func to go into a top level storage package. I just merged #248, which creates the `github.com/deis/builder/pkg/storage` package, so that's a perfect place to put it

pkg/gitreceive/k8s_util.go
```diff
@@ -15,10 +15,10 @@ const (
 	slugBuilderName   = "deis-slugbuilder"
 	dockerBuilderName = "deis-dockerbuilder"
 
-	tarURLKey        = "TAR_URL"
-	putURLKey        = "put_url"
+	tarPath          = "TAR_PATH"
+	putPath          = "PUT_PATH"
```
arschles: @kmala does the slugbuilder look for this new env variable?

pkg/gitreceive/storage/slug_builder_info.go
```diff
@@ -1,10 +1,6 @@
 package storage
 
-import (
-	"fmt"
-
-	"github.com/deis/builder/pkg/gitreceive/git"
-)
+import "fmt"
```
arschles: @kmala can you still put imports in parentheses, even if there is one? This one would be the following:

```go
import (
	"fmt"
)
```

pkg/gitreceive/storage/slug_builder_info.go
```diff
@@ -14,23 +10,18 @@ const (
...
-// NewSlugBuilderInfo creates and populates a new SlugBuilderInfo based on the given data.
-func NewSlugBuilderInfo(s3Endpoint *Endpoint, bucket, appName, slugName string, gitSha *git.SHA) *SlugBuilderInfo {
+// NewSlugBuilderInfo creates and populates a new SlugBuilderInfo based on the given data
+func NewSlugBuilderInfo(slugName string) *SlugBuilderInfo {
 	tarKey := fmt.Sprintf("home/%s/tar", slugName)
-	// this is where the controller tells slugrunner to download the slug from, so we have to
-	// tell slugbuilder to upload it to here
-	pushKey := fmt.Sprintf("home/%s:git-%s/push", appName, gitSha.Short())
+	// this is where workflow tells slugrunner to download the slug from, so we have to tell slugbuilder to upload it to here
+	pushKey := fmt.Sprintf("home/%s/push", slugName)
```
arschles: why this change? the push key was originally constructed so anyone could identify the app name and the git sha from looking at it

glide.yaml
kmala: I would be using that once i get this merged https://github.com/deis/distribution/pull/7/files

pkg/gitreceive/k8s_util.go
kmala: Yes.

glide.yaml
arschles: ok, thanks for the pointer. I just gave a review there. I'll add that PR to the required list for this one.

arschles:
Requires https://github.com/deis/distribution/pull/7/files

pkg/gitreceive/storage/slug_builder_info.go
kmala: slugName is constructed using appname and gitSha. So removed since its just duplication of code.
https://github.com/deis/builder/blob/master/pkg/gitreceive/build.go#L56

pkg/conf/config.go
kmala: I kept in conf because i am getting the storage params from the environment configuration.

arschles:
doesn't this require https://github.com/deis/slugbuilder/pull/41 and https://github.com/deis/dockerbuilder/pull/41?

kmala:
Yes. all the PR's in the first point https://github.com/deis/deis/issues/4966 except the Postgres should go together.

glide.yaml
```diff
@@ -32,4 +32,6 @@ import:
   version: ~1.1
 - package: github.com/arschles/assert
   version: 6882f85ccdc7c1822b146d1a6b0c2c48f91b5140
-- package: github.com/minio/minio-go
+- package: github.com/docker/distribution
```
arschles: Since https://github.com/docker/distribution/pull/1512 has been merged, can we now use the docker repo?

pkg/conf/config.go
```diff
@@ -35,3 +41,35 @@ func GetBuilderKey() (string, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, file := range files {
+		data, err := ioutil.ReadFile(storageCredLocation + file.Name())
+		if err != nil {
+			return nil, err
+		}
+		if file.Name() == "key.json" {
```
arschles: can you please put `key.json` into a constant

arschles: and also add a comment here explaining why this special case is necessary

pkg/conf/config.go
arschles: Ok, I see what you mean. I think the argument could be made either way, so I won't contest your decision.

pkg/gitreceive/build.go
```diff
@@ -16,8 +16,9 @@ import (
 	"github.com/deis/builder/pkg/storage"
 	"github.com/deis/builder/pkg/sys"
 	"github.com/deis/pkg/log"
+	"github.com/docker/distribution/context"
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
```
arschles: why not just leave this called `driver`?

pkg/gitreceive/build.go
```diff
@@ -100,21 +101,16 @@ func build(conf *Config, s3Client *storage.Client, kubeClient *client.Client, fs
...
 	if err != nil {
-		return fmt.Errorf("opening %s for read (%s)", appTgz, err)
+		return fmt.Errorf("error while reading file %s: (%s)", appTgz, err)
 	}
 
-	log.Debug("Uploading tar to %s/%s/%s", s3Client.Endpoint.FullURL(), conf.Bucket, slugBuilderInfo.TarKey())
-	if err := storage.UploadObject(s3Client, conf.Bucket, slugBuilderInfo.TarKey(), appTgzReader); err != nil {
-		return fmt.Errorf("uploading %s to %s/%s (%v)", absAppTgz, conf.Bucket, slugBuilderInfo.TarKey(), err)
-	}
+	//log.Debug("Uploading tar to %s/%s/%s", s3Client.Endpoint.FullURL(), conf.Bucket, slugBuilderInfo.TarKey())
```
arschles: can you either remove or uncomment?

pkg/gitreceive/build_test.go
```diff
@@ -7,6 +7,8 @@ import (
 	"os"
 	"testing"
 
+	"github.com/docker/distribution/context"
+
 	"gopkg.in/yaml.v2"
 
 	"github.com/arschles/assert"
```
arschles: can you remove the above two newlines?

glide.yaml
kmala: Docker distribution doesn't create a bucket if it isn't present and they have a regex which prevents us from using the ':' in the path https://github.com/docker/distribution/blob/master/registry/storage/driver/storagedriver.go#L109.
All our folders are <appname>:<gitsha>

pkg/gitreceive/build_test.go
kmala: I require the context in the code

glide.yaml
arschles: Ok, fine to leave as-is for now then.

Also, do you know the reasoning behind that regex? I'm particularly wondering if there's an object storage system that disallows `:` in object keys (I don't know of one)? Not blocking for this PR.

pkg/healthsrv/buckets_lister.go
```diff
@@ -1,19 +1,17 @@
 package healthsrv
 
-import (
-	s3 "github.com/minio/minio-go"
-)
+import "github.com/docker/distribution/context"
```
arschles: can you leave imports inside of parentheses?

pkg/healthsrv/buckets_lister.go
```diff
@@ -1,19 +1,17 @@
...
 
 // BucketLister is a *(github.com/minio/minio-go).Client compatible interface that provides just
```
arschles: Can you update this comment to reflect that it's now compatible with the storage driver?

Also, perhaps for a later PR, but did you consider using the [in memory storage driver](https://godoc.org/github.com/deis/distribution/registry/storage/driver/inmemory)?

pkg/storage/interfaces.go
```diff
@@ -1,187 +1,56 @@
...
-	// This function is the last one we have to add to make this interface have all the same
-	// functions as s3.Object.
-	Stat() (s3.ObjectInfo, error)
-}
-
-// ObjectGetter is the interface to get an object from object storage. The minio client doesn't
-// already satisfy this interface, because the GetObject func
-// (https://godoc.org/github.com/minio/minio-go#Client.GetObject) doesn't return an Object.
-// Instead, it returns a *s3.Object. Use the RealObjectGetter below to use the minio client.
+// ObjectGetter is a *(github.com/docker/distribution/registry/storage/driver/factory).StorageDriverFactory compatible interface, restricted to
```
arschles: Isn't this a `*(github.com/docker/distribution/registry/storage/driver).StorageDriver` compatible interface?

pkg/gitreceive/build.go
kmala: we can but i thought it is better to maintain consistency across all the repos.

pkg/gitreceive/build.go
```diff
@@ -211,17 +206,15 @@ func build(conf *Config, s3Client *storage.Client, kubeClient *client.Client, fs
...
 		conf.ObjectStorageWaitDuration(),
-		conf.Bucket,
 		slugBuilderInfo.AbsoluteSlugObjectKey(),
 	)
 	// poll the s3 server to ensure the slug exists
 	if !usingDockerfile {
 		if err := storage.WaitForObject(
-			s3Client,
-			conf.Bucket,
+			storageDriver,
```
smothiki: Now we are using cli libraries to access object storage do we have to do this check 
@arschles @kmala

pkg/gitreceive/build.go
arschles: ok, that's fine

pkg/gitreceive/build.go
arschles: @smothiki didn't we always do this check?

pkg/gitreceive/build_test.go
arschles: What context? It's just 2 newlines.

pkg/gitreceive/k8s_util.go
arschles: It seems like the slugbuilder will start looking for this new env var after https://github.com/deis/slugbuilder/pull/41. That's been marked on this PR as required for merge

pkg/gitreceive/k8s_util.go
```diff
@@ -63,14 +60,15 @@ func dockerBuilderPod(debug, withAuth bool, name, namespace string, env map[stri
...
+	pod := buildPod(debug, name, namespace, env)
 
 	pod.Spec.Containers[0].Name = slugBuilderName
 	pod.Spec.Containers[0].Image = slugBuilderImage
 
-	addEnvToPod(pod, tarURLKey, tarURL)
-	addEnvToPod(pod, putURLKey, putURL)
+	addEnvToPod(pod, tarPath, tarKey)
+	addEnvToPod(pod, putPath, putKey)
+	addEnvToPod(pod, "BUILDER_STORAGE", storageType)
```
arschles: can you please put `BUILDER_STORAGE` in a constant

pkg/gitreceive/k8s_util.go
```diff
@@ -99,23 +97,21 @@ func buildPod(debug, withAuth bool, name, namespace string, env map[string]inter
...
-			api.VolumeMount{
-				Name:      minioUser,
-				MountPath: "/var/run/secrets/object/store",
-				ReadOnly:  true,
-			},
-		}
+	pod.Spec.Containers[0].VolumeMounts = []api.VolumeMount{
+		api.VolumeMount{
+			Name:      objectStore,
+			MountPath: "/var/run/secrets/deis/objectstore/creds",
```
arschles: can you please put this path into a constant

pkg/gitreceive/build.go
smothiki: We used to do that for minio. I dont know why we did back then. But now its a python script for slugbuilder or a binary in future if the upload fails slugbuilder pod will exit with non zero status code, which we are already checking

pkg/gitreceive/run.go
```diff
@@ -6,11 +6,10 @@ import (
 	"os"
 	"strings"
 
-	"github.com/deis/builder/pkg/storage"
+	builderconf "github.com/deis/builder/pkg/conf"
```
arschles: why alias this to `builderconf`?

kmala: not sure...its there from the beginning...the imports got just restructured

arschles:
@kmala code LGTM.

Has this been manually tested with https://github.com/deis/slugbuilder/pull/41, https://github.com/deis/slugrunner/pull/24 and https://github.com/deis/dockerbuilder/pull/41? It seems that all of those will need to be merged at the same time.

kmala:
Yes i had manually tested and ran the e2e tests locally...i will run them again before merging.

arschles:
@kmala another comment that I just thought of - there is [code in builder](https://github.com/deis/builder/blob/master/pkg/gitreceive/build.go#L103) to create a bucket for storing code (and later, slugs), but I see that particular code block got removed here. Is that functionality replaced anywhere else?

kmala:
@arschles yes its moved to the storage layer https://github.com/deis/distribution/blob/master/registry/storage/driver/s3-aws/s3.go#L308

PR #270, opened 2016-03-24
feat(gitreceive): remove tmp build dir on success

bacongobbler:
port of deis/deis#4990

### DIFF
--- a/pkg/gitreceive/build.go
+++ b/pkg/gitreceive/build.go
@@ -82,6 +82,11 @@ func build(
 	if err != nil {
 		return fmt.Errorf("unable to create tmpdir %s (%s)", buildDir, err)
 	}
+	defer func() {
+		if err := os.RemoveAll(tmpDir); err != nil {
+			fmt.Errorf("unable to remove tmpdir %s (%s)", tmpDir, err)
+		}
+	}()
 
 	slugBuilderInfo := NewSlugBuilderInfo(slugName)
 


arschles:
As of this writing, the temp dir is created at https://github.com/deis/builder/blob/master/pkg/gitreceive/build.go#L61

arschles:
Ignore my previous comment. I mixed this up with the corresponding issue. Clearly you know where the temp dir is created....

pkg/gitreceive/build.go
```diff
@@ -261,6 +261,10 @@ func build(conf *Config, s3Client *storage.Client, kubeClient *client.Client, fs
 		return fmt.Errorf("cleaning up the repository with %s (%s)", strings.Join(gcCmd.Args, " "), err)
 	}
 
+	if os.RemoveAll(tmpDir) != nil {
```
arschles: shouldn't this be `defer`-ed

arschles:
Is this a fix for https://github.com/deis/builder/issues/271?

bacongobbler:
I saw that. I guess you mean to implement #271 in this PR as well? It should be no problem to change it to `defer os.RemoveAll(tmpDir)` right after that line, but then we're silently ignoring the error.

arschles:
We could still log if the temp dir is not cleaned up. I think if the temp dir is not properly deleted, it's something operators would want to know about, but it shouldn't fail the build.

Sidenote: currently, the [`git gc` command](https://github.com/deis/builder/blob/master/pkg/gitreceive/build.go#L259-L262) fails the build, but it could be argued that a failure on this command shouldn't fail the build either.

bacongobbler:
@arschles I decided to go ahead and `defer` the call. It's still the more idiomatic way to go.

> Sidenote: currently, the git gc command fails the build, but it could be argued that a failure on this command shouldn't fail the build either.

Agreed. I'll file an issue for that.

bacongobbler:
#272 for the `git gc` bug

arschles:
thanks @bacongobbler !

arschles:
RE the build failure - downloading vendored deps failed :(

pkg/gitreceive/build.go
```diff
@@ -62,6 +62,7 @@ func build(conf *Config, s3Client *storage.Client, kubeClient *client.Client, fs
 	if err != nil {
 		return fmt.Errorf("unable to create tmpdir %s (%s)", buildDir, err)
 	}
+	defer os.RemoveAll(tmpDir)
```
arschles: could you log an error if the removal failed?

bacongobbler:
we can let this one hang for RC1 so I don't mind that it's going red for now. It appears that bitbucket is under maintenance though :(

arschles:
@bacongobbler sgtm

pkg/gitreceive/build.go
bacongobbler: Would this piece of code work? I suck at writing closures

```
defer func() {
    if err := os.RemoveAll(tmpDir); err != nil {
        fmt.Errorf("unable to remove tmpdir %s (%s)", tmpDir, err)
    }
}()
```

arschles: Yep, that's perfect.

pkg/gitreceive/build.go
```diff
@@ -82,6 +82,11 @@ func build(
 	if err != nil {
 		return fmt.Errorf("unable to create tmpdir %s (%s)", buildDir, err)
 	}
+	defer func() {
+		if err := os.RemoveAll(tmpDir); err != nil {
+			fmt.Errorf("unable to create tmpdir %s (%s)", tmpDir, err)
```
arschles: @bacongobbler this log should probably say "unable to remove tmpdir..."

bacongobbler: oops, nice catch

arschles:
@bacongobbler with the exception of my small [comment](https://github.com/deis/builder/pull/270#discussion_r60967872), code LGTM

Issue #96, opened 2016-01-14
convert USING_DOCKERFILE to boolean


bacongobbler:
see https://github.com/deis/controller/issues/278

vdice:
@arschles is this ticket still in progress and/or still intended for inclusion in the `v2.0-rc1` release?

arschles:
no, moving out of a milestone completely, since #276 was likewise moved

PR #276, opened 2016-03-29
fix(pkg/types.go): change Dockerfile field to a bool

arschles:
Fixes https://github.com/deis/builder/issues/96
Requires https://github.com/deis/controller/issues/278

cc/ @helgi

### DIFF
--- a/pkg/gitreceive/controller.go
+++ b/pkg/gitreceive/controller.go
@@ -166,11 +166,11 @@ func createBuildHook(
 		Procfile:    procType,
 	}
 	if !usingDockerfile {
-		ret.Dockerfile = ""
+		ret.Dockerfile = false
 		// need this to tell the controller what URL to give the slug runner
 		ret.Image = slugBuilderInfo.AbsoluteSlugObjectKey()
 	} else {
-		ret.Dockerfile = "true"
+		ret.Dockerfile = true
 	}
 	return ret
 }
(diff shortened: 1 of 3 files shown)


arschles:
@helgi should I still pursue this change? Not sure what the controller expects. Regardless, moving to RC1 for now

mboersma:
This needs a rebase and should be merged for rc1 or probably just closed.ping @arschles @helgi

helgi:
@mboersma There is no corresponding change for the Controller. I spent a bit of time before on this and the migration path wasn't one I could get to work. Do you want to try to do it? :P

mboersma:
@helgi this just seemed like a potential non-backward-compatible change that I wanted to resolve before rc1. The original justification in deis/controller#278 doesn't sound urgent:
> This field when blank is considered false, and true when non-blank. We should just change it to a boolean now that we're breaking APIs.

Unless @bacongobbler thinks the blank/non-blank boolean behavior is potentially buggy, and we can safely make the change for rc1, I think we should close this and the originating issue.

helgi:
@mboersma it was more a "now is the time to break API if ever there was one"

arschles:
Anyone have a consensus on what to do here? FWIW since this needs a rebase and we also need to implement the corresponding change in the controller, my opinion is this patch is more trouble than it's worth.

bacongobbler:
Very low priority to fix. It's more of a nitpick since this behaviour of true/false in a string works, just not optimal. If it's too much of a pain to fix then we can come back to this in v3

arschles:
ok, moving this out of a milestone completely. can revisit after v2 is released

Joshua-Anderson:
I like this change, however, it would now need to happen in the controller SDK. @arschles Can this be closed?

I'm still in favor of making it, and I think we could do it in a backward compatible way.

1. We migrate the field in the controller to a bool, and update the SDK to also use a bool.
2. The controller could by updated to accept anything truthy for the field, so it would accept either the string or the bool.

Alternatively, we could say that the hooks api is an internal api and is therefore not constrained by semver, so we could make backwards incompatible changes to it.

arschles:
@Joshua-Anderson yes, this can be closed and so can the issue IMO. I opened this PR for exemplary purposes only. Note that if we close this, we can (but don't have to) close #96 and its companion deis/controller#278

PR #294, opened 2016-04-11
fix(pkg/git/git.go): fix blank image pull policy

### LINKED ISSUE #293: Don't assign pull policy environment variables in code
# Summary

We have made pull policy configurable but introduced a tiny bug.

## If You Are Reporting a Bug

https://github.com/deis/builder/blob/master/pkg/git/git.go#L41
Assigning a pull policy environment variable to pullpolicy variable .
If the env is not set it will have null and env config doesn't fall to defaults.

### Replication Steps

1. Create a Deis Cluster
2. Register an app
3. git push deis master should fail with invalid pull policy error 

Also, please provide a description of the desired result after the tester completes the above steps.

1. The app called "abcd" should be deployed
2. `kubectl get pod --namespace=abcd` should show 2 pods running
3. etc...

arschles:
# Summary of Changes

This PR removes the environment variable declarations from the git pre-receive hook script. When any one of the environment variables is not set in the main process, it's sent into the pre-receive hook process as the empty string, and that was leading to configuration errors in the hook, which was failing builds.

For example, if `SLUG_BUILDER_IMAGE_PULL_POLICY` was empty in the main process, it would be sent to `""` in the hook before this change. Then, the hook would try to convert `""` to a valid pull policy, which would fail, and the entire build would fail.

Fixes #293

None

# Associated [Documentation](https://github.com/deis/docs-v2) PR(s)

None

None

# Testing Instructions

1. Start a builder with no `SLUG_BUILDER_IMAGE_PULL_POLICY` or `DOCKER_BUILDER_IMAGE_PULL_POLICY` image pull policy environment variables set
2. Do a `git push` with a buildpack app
3. Do a `git push` with a dockerfile app

The result of both steps 2 and 3 should be a successful build, and in both cases, the appropriate image (slugbuilder and dockerbuilder, respectively) should be pulled in all cases (i.e. with `imagePullPolicy: Always`)

# Pull Request Hygiene TODOs

### DIFF
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -31,7 +31,14 @@ strip_remote_prefix() {
     stdbuf -i0 -o0 -e0 sed "s/^/"$'\e[1G'"/"
 }
 
-GIT_HOME={{.GitHome}} boot git-receive | strip_remote_prefix
+GIT_HOME={{.GitHome}} \
+SSH_CONNECTION="$SSH_CONNECTION" \
+SSH_ORIGINAL_COMMAND="$SSH_ORIGINAL_COMMAND" \
+REPOSITORY="$RECEIVE_REPO" \
+USERNAME="$RECEIVE_USER" \
+FINGERPRINT="$RECEIVE_FINGERPRINT" \
+POD_NAMESPACE="$POD_NAMESPACE" \
+boot git-receive | strip_remote_prefix
 `
 
 var preReceiveHookTpl = template.Must(template.New("hooks").Parse(preReceiveHookTplStr))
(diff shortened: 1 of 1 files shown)


smothiki:
affects can be seen in this build https://ci.deis.io/job/workflow-test-pr/537/consoleFull

mboersma:
Thanks for fixing this--it's been making it very difficult to test a builder fix I'm working on.

helgi:
I'd recommend setting a default instead of removing

arschles:
@helgi defaults are set to `Always` [here](https://github.com/deis/builder/blob/master/pkg/sshd/config.go#L15-L16)

arschles:
Waiting for e2e tests to merge.

smothiki:
I have seen one more failure in the test might not be related to this PR but might require a restart

arschles:
@smothiki agree that it may not be related, but restarted

arschles:
The second test run has failed, and the failures look to be unrelated to this change (and this change has eliminated the failures that it intended to fix). Merging...

PR #308, opened 2016-04-25
fix(cleaner): handle the channel close gracefully

### LINKED ISSUE #287: High CPU usage on AWS cluster
Builder using high CPU on AWS clusters.
Replication:
Deploy a workflow-dev or workflow-beta1 charts and wait for about one hour, and then do a `docker stats` on the builder container:

    CONTAINER           CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O
    3846d98d2cc9        123.87%             22.88 MB / 3.958 GB   0.58%               0 B / 0 B           28.67 kB / 0 B

`top` on the node gives:
```
top - 21:54:23 up 14:31,  1 user,  load average: 2.68, 2.37, 2.28
Tasks: 114 total,   3 running, 103 sleeping,   0 stopped,   8 zombie
%Cpu(s): 58.7 us,  0.8 sy,  0.0 ni,  0.0 id,  0.0 wa,  0.0 hi,  0.0 si, 40.5 st
KiB Mem:   3865572 total,  3560500 used,   305072 free,   316248 buffers
KiB Swap:        0 total,        0 used,        0 free.  1861688 cached Mem

  PID USER      PR  NI    VIRT    RES    SHR S %CPU %MEM     TIME+ COMMAND
24647 root      20   0   24908  21756      4 S 88.1  0.6  45:22.33 boot
 3983 root      20   0  609148  60620  31492 S  3.9  1.6  17:17.51 kubelet
    3 root      20   0       0      0      0 R  1.0  0.0   1:13.18 ksoftirqd/0
    7 root      20   0       0      0      0 R  1.0  0.0   5:46.98 rcu_sched
   10 root      rt   0       0      0      0 S  1.0  0.0   0:08.62 watchdog/0
 5541 root      20   0 2057032 390208   8296 S  1.0 10.1  43:48.62 java
15137 root      20   0       0      0      0 S  1.0  0.0   0:00.37 kworker/0:0
```
as per `ps faux` i couldn't find any defunct child processes as below:

    root     24647 45.1  0.5  24908 21904 ?        Ssl  20:08  51:03  \_ boot server

This bug is probably related to #280 but PR #283 didn't apparently resolve it

For clarity, `quay.io/deisci/builder:git-434a078` was the image used by the latest available workflow-dev at the moment of this report.

Scaling down/up or deleting the sick builder pod will fix the problem for about another hour...

Also, I couldn't identify any issues within the builder logs...

kmala:
# Summary of Changes

when the channel is closed the break just comes out of the select statement and not the for loop which is causing it to be an empty for loop and hence the raise in the CPU usage.

Closes #287

# Testing Instructions

1. Create a Deis Cluster
2. Register an app to test builder is working
3. Now wait for ~1 hr so that the channel is closed which can be confirmed by seeing the builder logs.
4. You should not see any spike in the cpu usage.

# Pull Request Hygiene TODOs


boot.go
```diff
@@ -82,8 +82,12 @@ func main() {
 				log.Printf("Starting health check server on port %d", cnf.HealthSrvPort)
 				healthSrvCh := make(chan error)
 				go func() {
-					if err := healthsrv.Start(cnf.HealthSrvPort, kubeClient.Namespaces(), storageDriver, circ); err != nil {
-						healthSrvCh <- err
+					for{
```
arschles: @kmala can you put a space between `for` and `{`

boot.go
```diff
@@ -82,8 +82,12 @@ func main() {
...
+						if err := healthsrv.Start(cnf.HealthSrvPort, kubeClient.Namespaces(), storageDriver, circ); err != nil {
+							healthSrvCh <- err
+							break
+						}
+						log.Print("namespace watcher channel closed.Starting the watch again")
```
arschles: @kmala seems like it would be better to exit the process instead of trying to restart the health server. Thoughts?

pkg/cleaner/cleaner.go
```diff
@@ -24,12 +24,14 @@ func Run(gitHome string, nsLister k8s.NamespaceWatcher, fs sys.FS) error {
 	watcher, err := nsLister.Watch(nil, nil, "")
 	if err != nil {
 		log.Printf("unable to get watch events (%s)", err)
+		return err
```
arschles: @kmala is this change related to the above changes in `boot.go`?

arschles: same with the below  change

boot.go
bacongobbler: I wish go fmt had some custom tooling so you could enforce this :(

boot.go
kmala: that would mean we wouldn't be cleaning the apps until the builder has restarted.....currently on my cluster the channel is getting closed in less than an hour which would mean we would be cleaning only during the first hour

pkg/cleaner/cleaner.go
kmala: there is code in boot.go(https://github.com/deis/builder/blob/master/boot.go#L110-L112) which is handling the errors and restarting the container but we are never returning the error.

The below code is to restart the process if the channel has closed.

boot.go
arschles: @kmala the builder won't be serving `git push`es until it restarts anyway. am I missing something here?

pkg/cleaner/cleaner.go
arschles: got it, thanks

pkg/cleaner/cleaner.go
```diff
@@ -24,12 +24,14 @@ func Run(gitHome string, nsLister k8s.NamespaceWatcher, fs sys.FS) error {
...
 	}
+
 	for {
 		select {
 		case event, ok := <-watcher.ResultChan():
 			if !ok {
-				break
+				return nil
```
arschles: this seems like an error condition. can you return a custom error? i.e. `errors.New("watcher result channel has been closed")`

boot.go
kmala: @arschles i think i misunderstood your point....do you mean exiting of the builder process? i am not sure if there is need to restart the builder when the watcher closes its channel as it  doesn't effect any builder flow?

arschles: >do you mean exiting of the builder process

Yep

>it doesn't effect any builder flow?

Well, the builder won't function properly if the watcher dies (deleted apps won't be cleaned up), so I'd say it does have an effect on builder flow

kmala: since we are getting the connection again and starting cleaning of the apps i don't think it would disturb build flow as only when the disk space fills does it have a problem.
What if the connections regularly closes which means we would be restarting the pod regularly and  i think the pod should be restarted only when it is absolute necessity that only a restart will help it.

arschles: well, actually I take it back. I forgot that the `healthSrvCh <- err` line ends up terminating the process anyway.

arschles:
I'm punting this to beta4

kmala:
closing in favor of #310

PR #324, opened 2016-05-08
feat(controller):Show the error message from controller instead of response code

### LINKED ISSUE #323: Show the error message sent by controller when a call fails
# Summary

Currently when a call to the controller fails we simply just output the  response code and the actual error message.To know the actual reason for error we need to look at the controller logs.Instead builder should show the error message sent by the controller.

## If You Are Reporting a Bug

If you're reporting a bug, please provide a detailed description of the problem.

### Replication Steps

1. Create a Deis Cluster
2. Register a default proc type docker file app by not exposing any port
3. build hook of controller fails and the returns an error.
4. builder just shows the response code instead of the error

## Everything Else

# Related Issues

kmala:
# Summary of Changes

Show the error message returned from the controller instead of the response code

Closes #323 

# Testing Instructions

1. Create a Deis Cluster
2. Deploy a docker app without exposing any port and proc file so that the controller will raise an exception and the build fails.
3. Builder should output the error message instead of the response code

# Pull Request Hygiene TODOs

### DIFF
--- a/pkg/gitreceive/controller.go
+++ b/pkg/gitreceive/controller.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/gitreceive/controller.go
```diff
@@ -17,18 +18,16 @@ var (
...
-	endpoint string
-	expected int
-	actual   int
+type unexpectedControllerError struct {
+	errorMsg string
 }
 
-func newUnexpectedControllerStatusCode(endpoint string, expectedCode, actualCode int) unexpectedControllerStatusCode {
-	return unexpectedControllerStatusCode{endpoint: endpoint, expected: expectedCode, actual: actualCode}
+func unExpectedControllerError(errorMsg string) unexpectedControllerError {
```
arschles: can you call this `newUnexpectedControllerError`?

pkg/gitreceive/k8s_util.go
```diff
@@ -241,7 +241,6 @@ func progress(msg string, interval time.Duration) chan bool {
 				return
 			case <-tick:
 				fmt.Println(msg)
-			default:
```
arschles: what's the purpose of taking out this `default:`? Also, how is it related to this PR?

pkg/gitreceive/controller.go
kmala: yes.

pkg/gitreceive/k8s_util.go
kmala: there is select inside an empty for loop and if there is a default case doing nothing which means the for loop gets run continuously for no reason

arschles: got it. I agree this should be removed, but it's out of scope of this PR. can you submit a different PR for removing this?

arschles:
code LGTM assuming https://github.com/deis/builder/pull/324#discussion_r62528725 and https://github.com/deis/builder/pull/324#discussion_r62528567 are done

pkg/gitreceive/k8s_util.go
kmala: sure..

Issue #220, opened 2016-03-02
Handle error case when repository lock unlock fails


arschles:
One possible fix is that the lock can become a lease.

Ref https://github.com/deis/builder/pull/212#discussion_r54755134

arschles:
Moving to RC1

arschles:
Looks like https://github.com/deis/builder/commit/aa7defbf1eeaf2243aee1786783148d0ec955d56 accidentally marked this as fixed. Reopening

arschles:
bumping from RC1, as this is not critical for the RC

mboersma:
Was this fixed by #367?

arschles:
Yes. Closing

Issue #298, opened 2016-04-15
[Meta] Race condition for when builder sometimes misses build pod runs


arschles:
# Summary

In cases where the slugbulder or dockerbuilder pod launches, runs and finishes before the builder starts looking for it, the builder will hang forever.

## If You Are Reporting a Bug

Please see the description above. 

### Replication Steps

As this is a race condition, it is non-deterministic. It has been reported in a variety of different scenarios, but here are some tips to increase the likelihood that it arises:

1. Ensure that the slugbuilder or dockerbuilder pod (depending on the type of your app) is already pulled on all nodes in your cluster.
2. Push a repository that will build quickly
3. Use fast object storage for your k8s deployment. For example, if you're running on [GKE](https://cloud.google.com/container-engine/), use [GCS](https://cloud.google.com/storage/).

When you've met the above conditions, simply follow these steps:

1. Create a deis cluster
2. Create a deis app (`deis create myapp`)
3. `git push deis master`

Desired result:

The builder hangs. The output immediately before the hang looks similar to the following:

```console
Pod spec: {
  "metadata": {
    "name": "slugbuild-gotest-e96335c6-899b3d78",
    "namespace": "deis",
    "creationTimestamp": null,
    "labels": {
      "heritage": "slugbuild-gotest-e96335c6-899b3d78"
    }
  },
  "spec": {
    "volumes": [
      {
        "name": "objectstorage-keyfile",
        "secret": {
          "secretName": "objectstorage-keyfile"
        }
      }
    ],
    "containers": [
      {
        "name": "deis-slugbuilder",
        "image": "quay.io/deisci/slugbuilder:git-c55ef21",
        "env": [
          {
            "name": "DEBUG",
            "value": "1"
          },
          {
            "name": "TAR_PATH",
            "value": "home/gotest:git-e96335c6/tar"
          },
          {
            "name": "PUT_PATH",
            "value": "home/gotest:git-e96335c6/push"
          },
          {
            "name": "BUILDER_STORAGE",
            "value": "gcs"
          }
        ],
        "resources": {},
        "volumeMounts": [
          {
            "name": "objectstorage-keyfile",
            "readOnly": true,
            "mountPath": "/var/run/secrets/deis/objectstore/creds"
          }
        ],
        "imagePullPolicy": "IfNotPresent"
      }
    ],
    "restartPolicy": "Never",
    "serviceAccountName": ""
  },
  "status": {}
}
```

[...]

PR #339, opened 2016-05-12
fix(pkg/sshd): wrap git receive functionality in a locking utility fu…

### LINKED ISSUE #309: Failed git pushes make it impossible to do a second git push
# Summary

As described in https://github.com/deis/builder/issues/204, when a `git push` fails, a second git push will fail with this error message: `[error] Another git push is ongoing`

## If You Are Reporting a Bug

The builder holds a lock on a per-repository basis, which doesn't get released when the `git push` is interrupted or fails for any reason.

### Replication Steps

1. `git push` a repository to an existing app on your Deis beta2 cluster
2. Interrupt the `git push `(i.e. using ctrl+c) while it is ongoing
3. Try to `git push` again

When you do the second `git push` in step 3, you should see the above error message. You may see a slightly different one due to #217 

# Related Issues

- #217 may cause error messages to differ
- #204 gave rise to this issue
- #220 is slightly related because repository unlocks may fail even if they're issued on purpose

arschles:
# Summary of Changes

This patch wraps the logic that runs the git receive hook in a utility function which calls `defer pushLock.Unlock()` so that, even if a `git push` fails, the lock gets unlocked.

# Associated End to End Tests

Created https://github.com/deis/workflow-e2e/issues/207 to track the work required to test this new behavior

Closes #309
Note that this patch doesn't attempt to fix https://github.com/deis/builder/issues/220

# Testing Instructions

1. Create a Deis Cluster, specifying a dockerbuilder and slugbuilder image that doesn't exist (see [here](https://github.com/deis/charts/blob/master/workflow-dev/tpl/deis-builder-rc.yaml#L35-L38))
2. Register an app
3. Add a key to your app
4. `git push deis master` from a repository
5. Cancel the `git push` while it's running by pressing ctrl+c

After step 5, you should be able to do another `git push` without getting an error, but you may have to wait a few seconds.

# Pull Request Hygiene TODOs


arschles:
Tested with the above testing instructions and verified that `git push`es can be interrupted and subsequent `git push`es succeed

bacongobbler:
execllent cleanup. Has anyone else manually tested this? considering that it's covered by unit tests I'd be okay with merging this, but just wondering if anyone else has tested this. :)

arschles:
@bacongobbler I'd love for more testing, so if you or anyone else has a chance to, I'm happy to wait before merging this PR

bacongobbler:
I have a cluster here so I'll manually test this.

bacongobbler:
:(

```
><> git push deis master
Counting objects: 80, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (45/45), done.
Writing objects: 100% (80/80), 19.02 KiB | 0 bytes/s, done.
Total 80 (delta 30), reused 80 (delta 30)
Starting build... but first, coffee!
...
^C
><> git push deis master
fatal: remote error: Another git push is ongoing
><> git push deis master
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

bacongobbler:
Hold up... if I wait a little longer pushes work again.

arschles:
@bacongobbler how long did you wait until pushes worked again?

bacongobbler:
about 20 seconds or so. Not that long.

arschles:
Hm, ok. I think what happened was that the lock didn't release until after the builder pod had finished. Since the git-receive hook runs in a new process, we have no mechanism right now for notifying it that the SSH channel was closed. I've created https://github.com/deis/builder/issues/345 to track that work.

Also, my testing instructions were incorrect - they should have described a way to make the slugbuilder pod fail to complete. I will update them now

bacongobbler:
> they should have described a way to make the slugbuilder pod fail to complete.

I have a buildpack [just for that](https://github.com/bacongobbler/heroku-buildpack-epic-fail). :)

arschles:
Perfect - I actually tried with an invalid slugbuilder image name and the builder was able to recover and allow new `git push`es after a short period of time. However, I noticed that the git-receive hook has issues reliably detecting when a builder pod fails to run to completion (for example, when it goes into `ErrImagePull`) - possibly related to #298.

The hook appears to eventually fail out, and the delay might be because it polls the state of the builder pod on an interval. Anyway, the fact that this PR allows the builder to heal itself is progress at least ;)

PR #349, opened 2016-05-18
chore(packages): use github mirror for speter/inf

slack:


arschles:
@slack can you make the PR description match the contributor guidelines? Or, this can be done if/when you merge-and-squash

glide.lock
```diff
@@ -167,7 +160,7 @@ imports:
 - name: github.com/Sirupsen/logrus
   version: 55eb11d21d2a31a3cc93838241d04800f52e823d
 - name: github.com/spf13/pflag
-  version: 8e7dc108ab3a1ab6ce6d922bbaff5657b88e8e49
+  version: 5644820622454e71517561946e3d94b9f9db6842
```
kmala: i am not totally sure of this...in the past i had some issue with using a different version and hence we fixed to specific commit

kmala: so we still have the issues updating it(travis confirmed it)..can we revert back the other changes except the speter

arschles: @kmala is it pinned in the `glide.yaml`? if not, we should do it in this patch

kmala: i think i tried doing that and had some other issue because of which i had to change the glide.lock file....may be we can try again to pin in the glide.yaml and see

slack: Happy to pin! Figured the glide-up would cause issues :(

slack: Actually, should I pin everything? or just spf13/pflag?

slack:
Ok dokes, pinned and re-pushed.

PR #350, opened 2016-05-20
chore(Dockerfile): Refactor image to use ubuntu-slim

krancour:


arschles:
@krancour this LGTM. has it been tested in the wild, or does it need to be?

krancour:
@arschles it should run through the normal CI process.

arschles:
@krancour sounds good. any special corners that should be covered here, that CI may not cover?

krancour:
Not that I can think of.

arschles:
code LGTM, pending the jenkins CI run succeeding

PR #354, opened 2016-05-31
feat(lock): add timeout to repository lock feature

smothiki:
# Summary of Changes
This PR implements a lock timeout. Where you can do git push again if the on going git push timeouts 

# Testing Instructions

1. Create a Deis Cluster
2. Register an app
3. Your Steps Here...

Also, please provide a description of the desired result after the tester completes the above steps.

1. The app called "abcd" should be deployed
2. `kubectl get pod --namespace=abcd` should show 2 pods running
3. etc...

# Pull Request Hygiene TODOs

### DIFF
--- a/boot.go
+++ b/boot.go
@@ -53,7 +53,7 @@ func main() {
 				}
 				fs := sys.RealFS()
 				env := sys.RealEnv()
-				pushLock := sshd.NewInMemoryRepositoryLock()
+				pushLock := sshd.NewInMemoryRepositoryLock(cnf.GitLockTimeout())
 				circ := sshd.NewCircuit()
 
 				storageParams, err := conf.GetStorageParams(env)
--- a/pkg/sshd/config.go
+++ b/pkg/sshd/config.go
@@ -14,9 +14,15 @@ type Config struct {
 	StorageType                  string `envconfig:"BUILDER_STORAGE" default:"minio"`
 	SlugBuilderImagePullPolicy   string `envconfig:"SLUG_BUILDER_IMAGE_PULL_POLICY" default:"Always"`
 	DockerBuilderImagePullPolicy string `envconfig:"DOCKER_BUILDER_IMAGE_PULL_POLICY" default:"Always"`
+	LockTimeout                  int    `envconfig:"GIT_LOCK_TIMEOUT" default:"10"`
 }
 
 // CleanerPollSleepDuration returns c.CleanerPollSleepDurationSec as a time.Duration.
 func (c Config) CleanerPollSleepDuration() time.Duration {
 	return time.Duration(c.CleanerPollSleepDurationSec) * time.Second
+}
+
+//GitLockTimeout return LockTimeout in minutes
+func (c Config) GitLockTimeout() time.Duration {
+	return time.Duration(c.LockTimeout) * time.Minute
 }
--- a/pkg/sshd/lock.go
+++ b/pkg/sshd/lock.go
@@ -14,41 +14,57 @@ var (
 // RepositoryLock interface that allows the creation of a lock associated
 // with a repository name to avoid simultaneous git operations.
 type RepositoryLock interface {
-	// Lock acquires a lock for a repository. In the case the repository is already locked
-	// it waits until a timeout to get the lock. If it was not possible to get the
-	// lock after the timeout an error is returned.
-	Lock(repoName string, timeout time.Duration) error
+	// Lock acquires a lock for a repository.
+	Lock(repoName string) error
 	// Unlock releases the lock for a repository or returns an error if the specified
-	// name doesn't exist. In the case the repository is already locked it waits until
-	// a timeout to get the lock. If it was not possible to get the lock after the timeout
-	// an error is returned.
-	Unlock(repoName string, timeout time.Duration) error
+	// name doesn't exist.
+	Unlock(repoName string) error
+	// Timeout returns the time duration for which it has to hold the lock
+	Timeout() time.Duration
 }
 
-func wrapInLock(lck RepositoryLock, repoName string, timeout time.Duration, fn func() error) error {
-	if err := lck.Lock(repoName, timeout); err != nil {
+func wrapInLock(lck RepositoryLock, repoName string, fn func() error) error {
+	if err := lck.Lock(repoName); err != nil {
 		return errAlreadyLocked
 	}
-	defer lck.Unlock(repoName, timeout)
-	return fn()
+	timer := time.NewTimer(lck.Timeout())
+	defer timer.Stop()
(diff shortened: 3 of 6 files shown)


pkg/sshd/lock.go
```diff
@@ -17,7 +17,7 @@ type RepositoryLock interface {
 	// Lock acquires a lock for a repository. In the case the repository is already locked
 	// it waits until a timeout to get the lock. If it was not possible to get the
 	// lock after the timeout an error is returned.
-	Lock(repoName string, timeout time.Duration) error
```
arschles: @smothiki for some background, @aledbf originally added this timeout so that we could build distributed locks with timeouts. since we've decided not to support multiple builder instances for now, we don't need that functionality right now.

However, if you are removing this, can you remove the godoc above that references it?

pkg/sshd/lock.go
```diff
@@ -26,11 +26,32 @@ type RepositoryLock interface {
 }
```
arschles: similar to the above removal in `Lock`, should this timeout be removed?

pkg/sshd/lock.go
```diff
@@ -26,11 +26,32 @@ type RepositoryLock interface {
...
 
 func wrapInLock(lck RepositoryLock, repoName string, timeout time.Duration, fn func() error) error {
-	if err := lck.Lock(repoName, timeout); err != nil {
+	if err := lck.Lock(repoName); err != nil {
 		return errAlreadyLocked
 	}
-	defer lck.Unlock(repoName, timeout)
-	return fn()
+	tick := time.Tick(timeout)
```
arschles: the ticker created here will tick forever, which doesn't seem to be what you want and will leak. it looks instead like you should create a [Timer](https://godoc.org/time#Timer)

pkg/sshd/lock.go
smothiki: Sure

sgoings:
Removing rc2 assignment because rc2 will only contain wfm and https://github.com/deis/builder/pull/353

pkg/sshd/lock.go
```diff
@@ -26,11 +26,32 @@ type RepositoryLock interface {
...
+	var fnerr error
+	go func() {
+		fnerr = fn()
+		quit <- false
+	}()
+	go func() {
+		for {
+			select {
+			case <-tick:
+				quit <- true
```
arschles: this will block forever if the function succeeds and `quit <- true` (above) executes first

pkg/sshd/server.go
```diff
@@ -248,7 +248,7 @@ func (s *server) answer(channel ssh.Channel, requests <-chan *ssh.Request, conda
 					channel.Stderr().Write([]byte("No repo given"))
 					return err
 				}
-				wrapErr := wrapInLock(s.pushLock, repoName, time.Duration(0), s.runReceive(req, sshconn, channel, repoName, parts, condata))
+				wrapErr := wrapInLock(s.pushLock, repoName, time.Duration(10*time.Minute), s.runReceive(req, sshconn, channel, repoName, parts, condata))
```
arschles: can you make this duration configurable?

smothiki:
requires https://github.com/deis/charts/pull/278

boot.go
```diff
@@ -53,7 +54,8 @@ func main() {
 				}
 				fs := sys.RealFS()
 				env := sys.RealEnv()
-				pushLock := sshd.NewInMemoryRepositoryLock()
+				locktimeout, _ := strconv.Atoi(env.Get("GIT_LOCK_TIMEOUT"))
```
arschles: can you put this into the existing [config struct](https://github.com/deis/builder/blob/master/pkg/sshd/config.go)

smothiki: Sure . Sorry I missed that

pkg/sshd/lock.go
```diff
@@ -17,38 +17,56 @@ type RepositoryLock interface {
...
+	}()
+	go func() {
+		fnerr = fn()
+		quit <- timer.Stop()
+	}()
+	quitter := <-quit
+	lck.Unlock(repoName)
+	if !quitter {
+		return fmt.Errorf("%s lock exceeded timout", repoName)
+	}
```
arschles: I get what this is doing and generally how it works, but it took me a few minutes to figure out _why_, and I think the part that's easy to miss is the buffered channel.

I think to make this code easier to read, it would be helpful to:

1. Use a `select` to receive the timer channel in the same goroutine as `wrapInLock` executes
2. Use a `select` to cancel the goroutine that calls `fn()` if execution times out

I've written an example similar to what I'm referring to [here](https://play.golang.org/p/udDlToS1bJ). I think the reduction of the goroutine and removal of the buffered channel makes it easier to grasp. Let me know thoughts.

arschles:
@smothiki looks pretty good. two remaining items from me:

- Still needs godoc edit: https://github.com/deis/builder/pull/354#discussion_r65275568
- I'd like to hear your thoughts on modifying the timeout logic: https://github.com/deis/builder/pull/354#discussion_r65461158

pkg/sshd/config.go
```diff
@@ -14,9 +14,15 @@ type Config struct {
...
 
 // CleanerPollSleepDuration returns c.CleanerPollSleepDurationSec as a time.Duration.
 func (c Config) CleanerPollSleepDuration() time.Duration {
 	return time.Duration(c.CleanerPollSleepDurationSec) * time.Second
 }
+
+//GitLockTimeout return LockTimeout in minutes
+func (c Config) GitLockTimeout() time.Duration {
+	return time.Duration(c.LockTimeout) * time.Minute
+}
```
arschles: :+1:

pkg/sshd/lock.go
```diff
@@ -17,38 +17,58 @@ type RepositoryLock interface {
...
+		case <-doneCh:
+		}
+	}()
+	select {
+	case <-timer.C:
+		lck.Unlock(repoName)
+		return fmt.Errorf("%s lock exceeded timout", repoName)
+	case err := <-fnCh:
+		return err
+	}
```
arschles: :+1:

pkg/sshd/lock.go
```diff
@@ -17,38 +17,58 @@ type RepositoryLock interface {
...
+	go func() {
+		err := fn()
+		select {
+		case fnCh <- err:
+		case <-doneCh:
+		}
+	}()
+	select {
+	case <-timer.C:
+		lck.Unlock(repoName)
```
arschles: make sure to `close(doneCh)` here, so the goroutine doesn't block forever

arschles: nevermind, I missed the `defer close(doneCh)` above. sorry!

PR #355, opened 2016-06-02
ref(k8s): update k8s api to 1.2.4

### LINKED ISSUE #343: k8s api rate limit
# Summary

Updating the k8s dependencies to 1.2 I saw the next output from a build:
```
remote: ---> W0513 02:29:46.517144      26 request.go:627] Throttling request took 67.040356ms, request: GET:https://10.3.0.1:443/api/v1/namespaces/deis/pods/slugbuild-rest-cert-avaluo-fiscal-6d837cb4-ec811c90        
remote: W0513 02:29:46.717168      26 request.go:627] Throttling request took 167.090257ms, request: GET:https://10.3.0.1:443/api/v1/namespaces/deis/pods/slugbuild-rest-cert-avaluo-fiscal-6d837cb4-ec811c90        
remote: W0513 02:29:46.917105      26 request.go:627] Throttling request took 167.006894ms, request: GET:https://10.3.0.1:443/api/v1/namespaces/deis/pods/slugbuild-rest-cert-avaluo-fiscal-6d837cb4-ec811c90 
```

which means the call to wait for the pod is hitting the api server too often.

I think this https://github.com/deis/builder/blob/master/pkg/gitreceive/build.go#L186 function should be refactoring to use a local cache. Something like this:

```
var podLister  cache.StorePodLister

podEventHandler := framework.ResourceEventHandlerFuncs{
  AddFunc: func(obj interface{}) {
    // new build pod
  },
  DeleteFunc: func(obj interface{}) {
    // build pod ended/died
  },
}

podLister.Store, podController = framework.NewInformer(
  &cache.ListWatch{
    ListFunc:  podListFunc(kubeClient, namespace),
    WatchFunc: podWatchFunc(kubeClient, namespace),
  },
  &api.Pod{}, resyncPeriod, podEventHandler)


func podListFunc(c *client.Client, ns string) func(api.ListOptions) (runtime.Object, error) {
  return func(opts api.ListOptions) (runtime.Object, error) {
    // ad condition to only receiver pods related with build operations buildpack/dockerfile
    return c.Pods(ns).List(api.ListOptions{
    })
  }
}

func podWatchFunc(c *client.Client, ns string) func(api.ListOptions) (runtime.Object, error) {
  return func(opts api.ListOptions) (runtime.Object, error) {
    // ad condition to only receiver pods related with build operations buildpack/dockerfile
    return c.Pods(ns).Watch(api.ListOptions{
    })
  }
}
```

and then `waitForPod` can use `podLister` instead of direct calls to kube api server.
This will reduce dramatically the number of calls and remove the rate limit (there's no limit using a local cache)

### LINKED ISSUE #344: Error upgrading k8s dependencies (deis/distribution fault)
# Summary

In https://github.com/deis/builder/pull/337 I tried to update the go k8s dependencies to 1.2
I failed because:

`speter.net/go/exp/math/dec/inf` changed to https://gopkg.in/inf.v0
the distribution fork is old. Current master changed to vendor instead of godeps

To make it work is necessary:
-use upstream `docker/distribution`
- apply:
    - deis/distribution@bac2cc5
    - deis/distribution@79483aa

Use #337 glide.yaml

kmala:
# Summary of Changes

change the k8s api version to 1.2.4 and fix the wait for pod as k8s now logs throttling requests.

Please list the issue(s) that this PR closes, similar to the below:

Closes #343 
Closes #344 

# Testing Instructions

1. Create a Deis Cluster
2. deploy an app with different storage backends
3. app should be deployed without any error

# Pull Request Hygiene TODOs

### DIFF
--- a/pkg/gitreceive/build.go
+++ b/pkg/gitreceive/build.go
@@ -190,7 +190,7 @@ func build(
 	req := kubeClient.Get().Namespace(newPod.Namespace).Name(newPod.Name).Resource("pods").SubResource("log").VersionedParams(
 		&api.PodLogOptions{
 			Follow: true,
-		}, api.Scheme)
+		}, api.ParameterCodec)
 
 	rc, err := req.Stream()
 	if err != nil {
--- a/pkg/gitreceive/build.go
+++ b/pkg/gitreceive/build.go
@@ -183,7 +183,12 @@ func build(
 		return fmt.Errorf("creating builder pod (%s)", err)
 	}
 
-	if err := waitForPod(kubeClient, newPod.Namespace, newPod.Name, conf.SessionIdleInterval(), conf.BuilderPodTickDuration(), conf.BuilderPodWaitDuration()); err != nil {
+	pw := k8s.NewPodWatcher(kubeClient, "deis")
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+	go pw.Controller.Run(stopCh)
+
+	if err := waitForPod(pw, newPod.Namespace, newPod.Name, conf.SessionIdleInterval(), conf.BuilderPodTickDuration(), conf.BuilderPodWaitDuration()); err != nil {
 		return fmt.Errorf("watching events for builder pod startup (%s)", err)
 	}
 
@@ -213,7 +218,7 @@ func build(
 	)
 	// check the state and exit code of the build pod.
 	// if the code is not 0 return error
-	if err := waitForPodEnd(kubeClient, newPod.Namespace, newPod.Name, conf.BuilderPodTickDuration(), conf.BuilderPodWaitDuration()); err != nil {
+	if err := waitForPodEnd(pw, newPod.Namespace, newPod.Name, conf.BuilderPodTickDuration(), conf.BuilderPodWaitDuration()); err != nil {
 		return fmt.Errorf("error getting builder pod status (%s)", err)
 	}
 	log.Debug("Done")
--- a/pkg/gitreceive/k8s_util.go
+++ b/pkg/gitreceive/k8s_util.go
@@ -4,10 +4,10 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/deis/builder/pkg/k8s"
 	"github.com/pborman/uuid"
 	"k8s.io/kubernetes/pkg/api"
-	apierrs "k8s.io/kubernetes/pkg/api/errors"
-	client "k8s.io/kubernetes/pkg/client/unversioned"
+	"k8s.io/kubernetes/pkg/labels"
 	"k8s.io/kubernetes/pkg/util/wait"
 )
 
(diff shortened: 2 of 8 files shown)


pkg/gitreceive/k8s_util.go
```diff
@@ -204,21 +204,19 @@ func waitForPodEnd(c *client.Client, ns, podName string, interval, timeout time.
 		return false, nil
 	}
 
-	return waitForPodCondition(c, ns, podName, condition, interval, timeout)
+	return waitForPodCondition(pw, ns, podName, condition, interval, timeout)
 }
 
 // waitForPodCondition waits for a pod in state defined by a condition (func)
-func waitForPodCondition(c *client.Client, ns, podName string, condition func(pod *api.Pod) (bool, error),
+func waitForPodCondition(pw *k8s.PodWatcher, ns, podName string, condition func(pod *api.Pod) (bool, error),
```
arschles: can you adapt `NamespaceWatcher` so that `*k8s.PodWatcher` conforms to it, and then make this `pw NamespaceWatcher`?

arschles: actually, I misunderstood what you're doing here. ignore my previous comment.

smothiki:
cool PR hope this will solve all inconsistencies and  race conditions in launching pods and getting logs

pkg/gitreceive/build.go
```diff
@@ -183,14 +183,19 @@ func build(
 		return fmt.Errorf("creating builder pod (%s)", err)
 	}
 
-	if err := waitForPod(kubeClient, newPod.Namespace, newPod.Name, conf.SessionIdleInterval(), conf.BuilderPodTickDuration(), conf.BuilderPodWaitDuration()); err != nil {
+	pw := k8s.NewPodWatcher(kubeClient, "deis")
```
aledbf: @kmala this should be in boot.go to "share" the watcher and persist while builder is running

Something like
```
	kubeClient, err := kcl.NewInCluster()
	if err != nil {
		return fmt.Errorf("couldn't reach the api server (%s)", err)
	}

	pw := k8s.NewBuildPodWatcher(kubeClient, "deis")
	go pw.Controller.Run(wait.NeverStop)
```

and pass pw as parameter to `gitreceive.Run(`

kmala: Right now we just need it during the slugbuilder/dockerbuilder pod which happens only during a new build which is not very frequent i assume.....Do we want Controller to be running continuously even when we may not need it?

aledbf: That's when you see the benefit. Please try to build more than one application at the same time (checking the amount of packets against the api server)

kmala: I had done 2 git pushes simultaneously and observed the packets using [iptraf](http://iptraf.seul.org/2.2/manual.html) and could see only one packet sent per app

PR #368, opened 2016-06-16
fix(flag):Use DEIS_DEBUG instead DEBUG environment variable for debug settings

kmala:
# Summary of Changes

check for DEIS_DEBUG as this is the one set in the charts.

# Testing Instructions

1. Create a Deis Cluster with DEIS_DEBUG falg set to true in the charts
2. Register an app
3. you should see the debug output

# Pull Request Hygiene TODOs

- [ ] Your commits follow the [commit style guidelines](http://docs.deis.io/en/latest/contributing/standards/#commit-style-guide)

### DIFF
--- a/boot.go
+++ b/boot.go
@@ -34,7 +34,7 @@ func init() {
 }
 
 func main() {
-	if os.Getenv("DEBUG") == "true" {
+	if os.Getenv("DEIS_DEBUG") == "true" {
 		pkglog.DefaultLogger.SetDebug(true)
 		log.Printf("Running in debug mode")
 	}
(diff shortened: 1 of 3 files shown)


boot.go
```diff
@@ -34,7 +34,7 @@ func init() {
 }
 
 func main() {
-	if os.Getenv("DEBUG") == "true" {
+	if os.Getenv("DEIS_DEBUG") == "true" {
```
arschles: can this use `conf.Debug` (would have to be moved down to after the config is parsed)

kmala: we can do that if we want but that would mean writing the same code for two different commands and currently the `server` command doesn't care about the gitreceive/conf where our debug flag is

arschles: ah, ok. nvm then!

PR #335, opened 2016-05-12
feat(probes): add controller check to the readiness probe

kmala:
# Summary of Changes

add probe to check the status of the controller

# Issue(s) that this PR Closes

Closes #181 

# Associated [End To End Test](https://github.com/deis/workflow-e2e) PR(s)

# Associated [Documentation](https://github.com/deis/workflow) PR(s)

# Associated [Design Document](http://docs.deis.io/en/latest/contributing/design-documents)(s)

# Testing Instructions

Please provide a detailed list for how to test the changes in this PR.

1. Create a Deis Cluster
2. check the status of the builder and it should be serving
3. now stop the controller and check the status of the builder, it should still be up but not ready to serve requests

# Pull Request Hygiene TODOs

Please make sure the below checklist is complete.

- [x] Your pull request is concise and to the point (make another PR for refactoring nearby code)
- [x] Your commits are squashed into logical units of work
- [x] Your commits follow the [commit style guidelines](http://docs.deis.io/en/latest/contributing/standards/#commit-style-guide)

PR #374, opened 2016-06-20
ref(readiness_handler): don't double-check controller

mboersma:
# Summary of Changes

Removes the pass-through readiness check from builder to controller. 

While connectivity to controller is required for builder, in practice this check appears to fail even in some situations where the `deis` CLI and controller RC checks continue to succeed. I would argue that this check isn't strictly necessary and is exacerbating the general problem described in deis/workflow-e2e#215, which occurs when builder fails readiness checks and traffic stops being routed to it.

Our assumption is that behavior differs since it bypasses deis-router, so perhaps nginx is papering over connectivity issues that might otherwise bite us.

Refs deis/workflow-e2e#215

# Pull Request Hygiene TODOs

- [X] Your pull request is concise and to the point (make another PR for refactoring nearby code)
- [X] Your commits follow the [commit style guidelines](http://docs.deis.io/en/latest/contributing/standards/#commit-style-guide)


kmala:
You can may try if reverting the PR helps https://github.com/deis/builder/pull/335

mboersma:
> if reverting the PR helps

Yeah, I'll do that instead--missed a few things trying to do it by hand. ✋

PR #376, opened 2016-06-29
chore(rootfs): bump slugbuilder/dockerbuilder images to v2.1.0

### LINKED ISSUE #357: Update pkg/gitreceive/config.go w/ v2.0.0 tags
# Summary

Once the `v2.0.0` milestone is closed (v2.0.0 released), the following [values](https://github.com/deis/builder/blob/master/pkg/gitreceive/config.go#L36-L37) need to be updated to point to the appropriate Docker image tags.

bacongobbler:
relies on the v2.1.0 release images being published.

closes #357


helgi:
These defaults are a little dangerous in general :|

bacongobbler:
> These defaults are a little dangerous in general :|

Could you elaborate on what makes these dangerous? Just the fact that we're pinning the image names or something else?

helgi:
That we pin them and they get forgotten, because we update things up in the charts

bacongobbler:
going to punt this to v2.2. This is not the approach we want to take. Instead we're just going to make this envvar a requirement so then we only have to reference this in one place.

bacongobbler:
#377 fixes this

PR #382, opened 2016-07-06
fix(pkg/conf): skip file if directory

### LINKED ISSUE #379: builder crashes on kube-cluster after cluster VMs get rebooted
# Summary

v2.0.0 builder crashes on kube-cluster after cluster VMs get rebooted

## If You Are Reporting a Bug

After successful deis-workflow 2.0.0 install on [kube-cluster](https://github.com/TheNewNormal/kube-cluster-osx), builder there works just fine.

### Replication Steps

1. Boot kube-cluster
2. Install Workflow with 'install_deis' script, which install Workflow with the default minio support
3. check with `kd get po` and builder shows Running status.
4. Halt the Up via menu to start VMs again
5. 'kd get po` shows `deis-builder-kpd52            0/1       CrashLoopBackOff   18         17h`
6. `kd logs deis-builder-kpd52` shows
```
2016/07/01 11:48:13 Error getting storage parameters (read /var/run/secrets/deis/objectstore/creds/..7987_01_07_11_42_31.882873540: is a directory)
```

Running Workflow on kube-solo, there is no such problem with the builder

bacongobbler:
The following code assumes that the file is a file, not a directory.

fixes #379

### DIFF
--- a/pkg/conf/config.go
+++ b/pkg/conf/config.go
@@ -51,6 +51,9 @@ func GetStorageParams(env sys.Env) (Parameters, error) {
 	}
 
 	for _, file := range files {
+		if file.IsDir() {
+			continue
+		}
 		data, err := ioutil.ReadFile(storageCredLocation + file.Name())
 		if err != nil {
 			return nil, err
--- a/pkg/conf/config_test.go
+++ b/pkg/conf/config_test.go
@@ -1,0 +1,56 @@
+package conf
+
+import (
+	"io/ioutil"
+	"os"
+	"os/user"
+	"testing"
+
+	"github.com/deis/builder/pkg/sys"
+)
+
+func TestGetStorageParams(t *testing.T) {
+	usr, err := user.Current()
+	if err != nil {
+		t.Logf("could not retrieve current user: %v", err)
+		t.SkipNow()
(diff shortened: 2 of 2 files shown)


mboersma:
Good job adding unit tests. 💚

bacongobbler:
If I'm bugging everyone else to add tests, it's only fair! :P

bacongobbler:
My only concern with this code is that we are writing to a location that usually is not writable by regular users (/var/run/secrets/...) so the only way to test this is to run the unit tests as root (which occurs within the test container). That really sucks for users relying on running tests on their own machine...

mboersma:
> the only way to test this is to run the unit tests as root

Well given that's the current CI environment, I think it's ok. I wonder if this could call `t.Skip()` if it weren't root? Or maybe just add an early comment like `// NOTE: these tests must run as root or a user with write access to /var/run/secrets`.

bacongobbler:
Using t.Skip() would be best for now. :)

bacongobbler:
I'll also see if I can increase coverage there. ^

bacongobbler:
k, tests should skip if not being run as root. :)

PR #383, opened 2016-07-06
docs(github): simplify github templates

bacongobbler:
We've found that half of the engineering team either hates or doesn't even bother using the PR
templates. Refactoring it down to the bare minimum should suffice.


arschles:
@bacongobbler could we leave a small section that prompts contributors to think about documentation at least? the [`deisrel` PR template](https://github.com/deis/deisrel/blob/master/.github/PULL_REQUEST_TEMPLATE) might be a good model to follow. note that I have no data to back up any of these claims - just a guess

bacongobbler:
That's a good suggestion. I've been trying to hack on simplifying down the templates to the "bare minimum". Making a two-line template to reference related doc or e2e PRs would be simple enough.

bacongobbler:
okay, that should simplify things!

PR #389, opened 2016-07-20
feat(registry): Add support for external registry

kmala:


pkg/conf/config.go
```diff
@@ -16,6 +16,7 @@ const (
 	gcsKey              = "key.json"
 )
 
+// Parameters is map which contains storage params
```
mboersma: Nitpick: Effective Go recommends doc strings be full sentences. Could you reword this slightly?
`// Parameters is a map which contains storage params.`

pkg/gitreceive/build.go
```diff
@@ -138,8 +138,21 @@ func build(
...
+		registryLocation := conf.RegistryLocation
+		registryEnv := make(map[string]string)
+		if registryLocation != "on-cluster" {
+			registryEnv, err = getRegistryDetails(kubeClient, &image, registryLocation, conf.PodNamespace, conf.RegistrySecretPrefix)
+			if err != nil {
+				return fmt.Errorf("error getting private registry details %s", err)
+			}
+			image = image + ":git-" + gitSha.Short()
+		}
+		registryEnv["DEIS_REGISTRY_PROXY_PORT"] = conf.RegistryProxyPort
```
bacongobbler: Note that I removed this config in https://github.com/deis/builder/pull/400. Was my assumption that RegistryProxyPort is actually important? From looking at the dockerbuilder code I didn't see a single reference to DEIS_REGISTRY_PROXY_PORT

bacongobbler: I'm also curious to why we're using a second envvar to represent the registry port when we can just override DEIS_REGISTRY_SERVICE_PORT instead?

kmala: DEIS_REGISTRY_PROXY_PORT was the initial env which i used to pass the port of the proy service and we  reverted the dockerbuilder and controller pr but din't the builder as we thought we want to bring the feature anyway.
Now, i see you are updating the DEIS_REGISTRY_SERVICE_PORT/HOST itself.
So either you or me has to rebase depending on who merges first 😄

kmala: I thought using the DEIS_REGISTRY_SERVICE_PORT will confuse as it is not the actual port on which registry service is running.

bacongobbler: :horse_racing:

bacongobbler: > I thought using the DEIS_REGISTRY_SERVICE_PORT will confuse as it is not the actual port on which registry service is running.

But it is, from the perspective of the builder! It just happens to be an nginx instance that forwards requests to the *real* registry. It's not a proxy server. You can interact with it the exact same as you could with the real registry, so I didn't see a reason why we can't re-use the same environment variables already existing.

Issue #326, opened 2016-05-10
Use Private Registry with Dockerfile


arkkanoid:
Hi, 
I've some images pushed to Docker Hub private registry and I want to use it with a Dockerfile deployment.
This only works if I pull previously the image on each node, then the app is deployed succesfully.

It's possible to deploy Dockerfile-apps that contains images from a private registry (without pull previously the images)?

bacongobbler:
You can use `deis pull` with `deis regisry:set`, but builder support (which includes Dockerfile builds) is not currently available.

arkkanoid:
Thanks @bacongobbler. It would be nice to have this feature.

bacongobbler:
let's keep this issue open to track this feature. :)

kmala:
fixed by https://github.com/deis/dockerbuilder/pull/101

PR #392, opened 2016-07-20
feat(dockerbuilder): expose docker registry authentication credentials to dockerbuilder

Joshua-Anderson:
ref #326 

This exposes docker credentials set with `deis registry` as a json environmental variable to dockerbuilder.

I'm working on a PR for dockerbuilder to use this environmental variable to authenticate with docker registries to allow for private base images.

### DIFF
--- a/pkg/gitreceive/build.go
+++ b/pkg/gitreceive/build.go
@@ -11,6 +11,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"gopkg.in/yaml.v2"
+
 	"github.com/deis/builder/pkg"
 	"github.com/deis/builder/pkg/git"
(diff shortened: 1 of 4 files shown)


kmala:
@Joshua-Anderson this may need to hold for time being as i am currently done with the support for replacing the internal registry with private registry( #389 ) and we haven't finalized on our approach on how to support the private registry for individual app(the current implementation using `deis regsitry:set ` doesn't fully support private registries and we may not want to go ahead with the current approach).

Joshua-Anderson:
@kmala No problem, I just saw #326 and thought it could be quickly fixed. When deis support for private registries is further along I'll take a look again and see what changes are necessary.

PR #394, opened 2016-07-21
feat(makefile): add formating and linting tests

Joshua-Anderson:


pkg/gitreceive/build.go
```diff
@@ -84,7 +84,7 @@ func build(
 	}
 	defer func() {
 		if err := os.RemoveAll(tmpDir); err != nil {
-			fmt.Errorf("unable to remove tmpdir %s (%s)", tmpDir, err)
+			log.Info("unable to remove tmpdir %s (%s)", tmpDir, err)
```
Joshua-Anderson: `govet` caught this. `fmt.Errorf` does nothing here, so I decided to replace this with a log so at least the error is now logged.

bacongobbler: Good catch.

PR #395, opened 2016-07-21
ref(Dockerfile): install git from PPA

### LINKED ISSUE #375: change dockerfile to use git's ppa
see https://github.com/deis/builder/pull/364#issuecomment-227852513

Joshua-Anderson:
Fixes #375


rootfs/Dockerfile
```diff
@@ -7,8 +7,11 @@ RUN adduser --system \
 	--group \
 	git
 
-RUN apt-get update \
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E1DF1F24 \
+	&& echo "deb http://ppa.launchpad.net/git-core/ppa/ubuntu xenial main" >> /etc/apt/sources.list \
+	&& apt-get update \
 	&& apt-get install -y \
+		git \
```
bacongobbler: Any way we can pin the version here so we can manually update this ourselves?

Joshua-Anderson: We can pin the version, but apt repositories rarely store older versions of packages, so we would be forced to upgrade anyway

bacongobbler: that's true but then again the apt repositories in ubuntu weren't updated in a long time so we had no way of upgrading to 2.8.4 without pinning and installing from source before. Just more concerned about history repeating itself. Anyways still LGTM

PR #396, opened 2016-07-21
feat(makefile): add deploy command to deploy image to k8s

Joshua-Anderson:
This is stolen from the controller makefile. It makes it way easier to update k8s to use the images you just built.


Makefile
```diff
@@ -56,4 +63,9 @@ docker-build: build
 docker-push:
 	docker push ${IMAGE}
 
+deploy: docker-build docker-push
+	sed 's#\(image:\) .*#\1 $(IMAGE)#' /tmp/deis-$(SHORT_NAME) | kubectl apply --validate=true -f -
```
bacongobbler: FYI eventually we can replace this with `kubectl set image rc deis-$(SHORT_NAME)`, if we can't already.

Joshua-Anderson: At least in kubectl version v1.2.4, I get `Error: unknown command "set" for "kubectl"`

bacongobbler: yeah I think it's a relatively new feature and it only works for deployments, IIRC. Saw it at the last sig-apps meeting. Anyways, LGTM!

PR #402, opened 2016-08-04
feat(*): switch to using controller-sdk-go for controller interactions

Joshua-Anderson:
After adding hook support to the controller-sdk-go, this migrates the builder to use it

refs deis/controller-sdk-go#69


glide.yaml
```diff
@@ -41,3 +41,6 @@ import:
   version: 8e7dc108ab3a1ab6ce6d922bbaff5657b88e8e49
   repo: https://github.com/spf13/pflag
   vcs: git
+- package: github.com/deis/controller-sdk-go
+  repo: https://github.com/joshua-anderson/controller-sdk-go
```
kmala: this should be changed after https://github.com/deis/controller-sdk-go/pull/69 is merged

pkg/gitreceive/run.go
```diff
@@ -48,9 +50,16 @@ func Run(conf *Config, fs sys.FS, env sys.Env, storageDriver storagedriver.Stora
 
 		log.Debug("read [%s,%s,%s]", oldRev, newRev, refName)
 
-		if err := receive(conf, builderKey, newRev); err != nil {
+		client, err := controller.New()
```
kmala: can we create the client outside of this for loop and pass that to the build function instead of creating again in the build function?

Joshua-Anderson: Fixed! `build()` now takes a parameter for a `deis.Client`. I pass the client created in run to the build method.

pkg/controller/utils.go
```diff
@@ -1,105 +1,48 @@
...
-	req.Header.Add("X-Deis-Builder-Auth", builderKey)
+	return client, nil
+}
 
-	res, err := http.DefaultClient.Do(req)
-	if err != nil {
-		return nil, err
-	}
-	defer res.Body.Close()
+// CheckAPICompt checks for API compatibility errors and warns aboit them.
```
bacongobbler: typo: aboit should be about

pkg/controller/utils.go
```diff
@@ -1,105 +1,48 @@
...
+func CheckAPICompt(c *deis.Client, err error) error {
```
bacongobbler: nitpick: mind changing the namep to `CheckAPICompat`?

Joshua-Anderson:
@bacongobbler: Fixed!

PR #403, opened 2016-08-05
ref(log): Log the error message when the auth to controller fails

### LINKED ISSUE #385: Better error message when builder/controller API auth fails
If the authentication from the builder to the controller fails due to a wrong secret, `deis push` fails with `Permission denied` and the builder logs `[ERROR] Failed handshake: EOF` which does not give any indication of the actual error.

Instead the builder should log something like `[ERROR] API authentication to controller failed` which would be much more useful when debugging the problem.

I ran into this while debugging the problem described in deis/charts#303 where `helmc` generated a builder secret that began with a space and caused the API request to verify the ssh key between builder and controller to fail.

In order to find the cause for the above mentioned issue I had to read the source code for the ssh server, controller utils, the go ssh module docs and construct an API request using `curl` from inside the builder container…

The code that should be augmented is in sshd `server.AuthKey()` calling `controller.UserInfoFromKey()` [without logging anything if this call fails](https://github.com/deis/builder/blob/3e7a785db5402cbd343a5f688d6798ecfe658efc/pkg/sshd/server.go#L40).

This issue is related to #32 and probably others relating to the "Failed handshake: EOF" error.

kmala:
fixes #385


helgi:
Is it possible to test this? I see it is not covered by unit tests right now

Joshua-Anderson:
@kmala This is going to be replaced by #402

kmala:
@Joshua-Anderson i see that even you are not logging the message....i will close this if you add it in your PR

Joshua-Anderson:
@kmala I shouldn't need to because the controller-sdk-go should return an verbose error. The error message is logged, correct?

kmala:
We pass the auth as a parameter(https://github.com/deis/builder/blob/3e7a785db5402cbd343a5f688d6798ecfe658efc/pkg/sshd/server.go#L69-L73) to the ssh server config https://github.com/deis/builder/blob/3e7a785db5402cbd343a5f688d6798ecfe658efc/pkg/sshd/server.go#L152-L156 and golang ssh doesn't return the proper error during an auth failure...it just sends EOF

Joshua-Anderson:
Got it. I thought this was logging it to the builder logs, not to the ssh output. Will add to my PR.

Joshua-Anderson:
Added this warning to #402

PR #405, opened 2016-08-08
fix(config): change the cpu limit to string instead of int

kmala:
cpu limit has to be string and int.
TestCase:
```
deis limits:set -c cmd=125m
git push deis master
```
this will fail before the fix with error `Error running git receive hook [getting app config for classy-teamwork (json: cannot unmarshal string into Go value of type int)]` and should pass after the fix


Joshua-Anderson:
I believe that this is fixed in the [controller SDK](https://github.com/deis/controller-sdk-go/blob/master/api/config.go#L32). Do we want to merge this or simply merge #402

bacongobbler:
Let's go forward with #402 to fix this

kmala:
yeah #402 should fix this. Closing this in favor of it.

PR #407, opened 2016-08-11
ref(gitreceive): remove push hook

Joshua-Anderson:
It was used for authorization, but the keys hooks is now used for both authentication and authorization.


helgi:
We'd not have the push audit log anymore, not that it is exposed anywhere right now

Joshua-Anderson:
Right now the push audit log isn't exposed anywhere. Is it worth exposing the push audit log and keeping the hook or getting rid of the audit log and the hook?

helgi:
@gabrtv, @mboersma and @bacongobbler would have some context here

Overall I'm fine with removing the hook - A different audit log approach can also be introduced back in the future

bacongobbler:
> Is it worth exposing the push audit log and keeping the hook or getting rid of the audit log and the hook?

what exactly are you referring to when you mean "the push audit log"? Haven't heard that terminology used before.

Joshua-Anderson:
All this hook does is add the fields of the request to a table in the DB. None of our code actual uses the table though.

Joshua-Anderson:
This is [the push model](https://github.com/deis/controller/blob/master/rootfs/api/models/push.py)

bacongobbler:
ah I see. Yes, server side it's just validation. Nothing is done with the request.

smothiki:
@bacongobbler @mboersma I dont have much context with Push Hook can you guys take a quick look or we can punt this to the next sprint .

PR #408, opened 2016-08-15
fix(gitreceive): change target name to Workflow

bacongobbler:


bacongobbler:
Everything's green except for codecov all because I changed one name. Sometimes codecov sucks. lol

PR #411, opened 2016-08-18
chore(Makefile): update to go-dev 0.17.0

mboersma:
Includes the go 1.7 toolchain: faster, smaller, better. See https://tip.golang.org/doc/go1.7.

Also updates the Docker base image to [deis/base:0.3.1](https://github.com/deis/docker-base/releases/tag/0.3.1).


mboersma:
The new approach to linting found a misspelling that wasn't previously flagged. That's awesome--thanks @Joshua-Anderson. I'll fix it up now.

Joshua-Anderson:
@mboersma We can always move it from mandatory to optional. I stuck it mandatory cause I hadn't seen any false positives yet. I'm looking to see if there is a way to ignore a line.

mboersma:
@Joshua-Anderson or we could just regenerate the PK to work around it. I'm a big fan of strict linting. :-)

PR #422, opened 2016-09-06
feat(builder): delete cache if the cache is disabled

Joshua-Anderson:
cc @jeroenvisser101

If `DEIS_DISABLE_CACHE` is set and the cache file exists, delete the cache.


pkg/gitreceive/build.go
```diff
@@ -112,6 +110,20 @@ func build(
 		}
 	}
 
+	_, disableCaching := appConf.Values["DEIS_DISABLE_CACHE"]
+	slugBuilderInfo := NewSlugBuilderInfo(appName, gitSha.Short(), disableCaching)
+
+	if slugBuilderInfo.DisableCaching() {
+		log.Debug("caching disabled for app %s", appName)
+		// If cache file exists, delete it
+		if _, err := storageDriver.Stat(context.Background(), slugBuilderInfo.CacheKey()); err == nil {
```
jeroenvisser101: Not sure, but I think these operations cost money, although I might be wrong.

pkg/gitreceive/build.go
```diff
@@ -178,14 +190,19 @@ func build(
 		)
 	} else {
 		buildPodName = slugBuilderPodName(appName, gitSha.Short())
+
+		cacheKey := ""
+		if !slugBuilderInfo.DisableCaching() {
+			cacheKey = slugBuilderInfo.CacheKey()
+		}
```
jeroenvisser101: Why did you add this exactly? When the cache is disabled, the cache key should still be a correct 'cache key', right? It should just be disabled, which it is.

jeroenvisser101:
Few remarks, but overal, this LGTM.

pkg/gitreceive/build.go
```diff
@@ -112,6 +110,20 @@ func build(
 		}
 	}
 
+	_, disableCaching := appConf.Values["DEIS_DISABLE_CACHE"]
```
kmala: this would mean user has to do config:set/unset DEIS_DISABLE_CACHE to disable/enable the cache.
Can we check the value instead?

jeroenvisser101: iirc, the application is now checking if it's 'set', which would mean this logic is correct. We could still change it though, although having `DEIS_DISABLE_CACHE=0` makes no sense to me and I'd rather see it being unset.

kmala: i think having `DEIS_DISABLE_CACHE=true/false` is better than set/unset

jeroenvisser101: I think my preference would be to do the same as `DEIS_KUBERNETES_DEPLOYMENTS` (I'm not sure how it's done there, but if we do it the same, we're consistent). Also, if you feel strongly about this, we might want to change both.

pkg/gitreceive/build.go
Joshua-Anderson: I removed some logic in `slugbuilderPod`, so it only adds `cacheKey` to the env if it's non-empty, instead of checking the map.

jeroenvisser101: ah, seems alright then 👍

pkg/gitreceive/build.go
Joshua-Anderson: They do cost money, but it's `$0.01 / 10K` per read API request on s3 at least, so I'd consider it negligible.

So it's much cheaper to check before attempting to delete, which costs `$0.01 / 1K`

Joshua-Anderson: I was incorrect, deleting a object on S3 is free. That being said, I suspect it's better practice to check if the file exists *before* trying to delete it, but that's me.

pkg/gitreceive/build.go
Joshua-Anderson: `DEIS_KUBERNETES_DEPLOYMENTS` worked similarly to this, but it's no longer in use, deployments having become the default.

pkg/gitreceive/build.go
jeroenvisser101: > They do cost money, but it's $0.004 / 10K per read API request on s3 at least, so I'd consider it negligible.

Totally, that's less than 1 cent per 20k builds, seems okay to me.

pkg/gitreceive/build.go
jeroenvisser101: Any other uses of these feature flags? Any pros/cons of using true/false vs. set/unset?

mboersma:
Set/unset is ok with me, since we did create a precedent for it with the deployments feature flag.

PR #424, opened 2016-09-08
fix(Makefile): update deploy target to use deployments

Joshua-Anderson:
Also clean up the makefile in general

This is based off the controller makefile.


Makefile
```diff
@@ -55,21 +45,17 @@ test-unit:
...
+	@if [ -z $$(which kubectl) ]; then \
+		echo "kubectl binary could not be located"; \
+		exit 2; \
+	fi
 
-deploy: docker-build docker-push
-	sed 's#\(image:\) .*#\1 $(IMAGE)#' /tmp/deis-$(SHORT_NAME) | kubectl apply --validate=true -f -
-	kubectl scale rc deis-$(SHORT_NAME) --replicas 0 --namespace deis
-	kubectl scale rc deis-$(SHORT_NAME) --replicas $(DESIRED_REPLICAS) --namespace deis
+deploy: check-kubectl docker-build docker-push
```
vdice: this still depends on the removed `docker-push` target

Joshua-Anderson: Nope, the `docker-push` target comes from `versioning.mk`. I'm not sure why this duplicated target is here.

vdice: n/m 👍  https://github.com/deis/builder/blob/master/versioning.mk#L15

PR #431, opened 2016-09-26
ref(controller): use the env config instead of directly getting from environment

### LINKED ISSUE #338: refactor the controller utils to the controller env variables
# Summary

currently controller utils(https://github.com/deis/builder/blob/master/pkg/controller/utils.go#L30)  directly gets the values of the environment values instead we should get it through the global configuration file we have(https://github.com/deis/builder/blob/master/pkg/conf/config.go#L29)

## Everything Else
# Related Issues
ref https://github.com/deis/builder/pull/335#discussion_r63078027

kmala:
fixes #338

### DIFF
--- a/pkg/conf/config.go
+++ b/pkg/conf/config.go
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	builderKeyLocation  = "/var/run/secrets/api/auth/builder-key"
+	BuilderKeyLocation  = "/var/run/secrets/api/auth/builder-key"
 	storageCredLocation = "/var/run/secrets/deis/objectstore/creds/"
 	minioHostEnvVar     = "DEIS_MINIO_SERVICE_HOST"
 	minioPortEnvVar     = "DEIS_MINIO_SERVICE_PORT"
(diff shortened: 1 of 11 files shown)


bacongobbler:
can you try to write unit tests for this if possible? Trying to increase coverage in #437 so it'd be helpful to have this all covered before merging.

kmala:
@bacongobbler i had added test.Can you review it again.

pkg/conf/config.go
```diff
@@ -36,10 +36,10 @@ func EnvConfig(appName string, conf interface{}) error {
 }
 
 // GetBuilderKey returns the key to be used as token to interact with deis-controller
-func GetBuilderKey() (string, error) {
-	builderKeyBytes, err := ioutil.ReadFile(builderKeyLocation)
+func GetBuilderKey(builderKeyPath string) (string, error) {
```
bacongobbler: Would it be easier to change BuilderKeyLocation to type `var` so tests can change the path before running `GetBuilderKey()`? Wondering why we'd define a constant `BuilderKeyLocation` variable, yet not use it for `GetBuilderKey()`.

bacongobbler: FYI I'm running into this same problem in #437 as well.

kmala: Just to be on the safeside that it we don't change it by mistake in other part of the code.I am sure that doesn't happen so i am ok with changing it to `var`

bacongobbler:
LGTM even without the change. This is partially blocking #437 as it touches on some of the code I'm directly testing in build.go (in a good way) :+1:

PR #445, opened 2016-10-23
feat(cleaner): delete from object store

### LINKED ISSUE #214: cleaner should also delete objects from S3 endpoint
RIght now cleaner only does deletion of App directories if an APP is deleted .
It should also clean objects from S3 endpoints

Joshua-Anderson:
The cleaner now deletes slug and cache files from object storage as well as locally. I didn't write any tests I'm not sure the best way to attempt to mock the object store driver and the cleaner function is untested.

I'm happy to write tests if you have pointers to ways to go about mocking the storage driver.

Fixes #214

### DIFF
--- a/pkg/gitreceive/slug_builder_info.go
+++ b/pkg/gitreceive/slug_builder_info.go
@@ -5,7 +5,9 @@ import (
 )
 
 const (
-	slugTGZName = "slug.tgz"
+	slugTGZName     = "slug.tgz"
+	CacheKeyPattern = "home/%s/cache"
+	GitKeyPattern   = "home/%s:git-%s"
 )
 
 // SlugBuilderInfo contains all of the object storage related information needed to pass to a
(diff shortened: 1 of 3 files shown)


Joshua-Anderson:
Rebased this, anything else needed to get this merged?

pkg/gitreceive/slug_builder_info.go
```diff
@@ -19,12 +21,12 @@ type SlugBuilderInfo struct {
 
 // NewSlugBuilderInfo creates and populates a new SlugBuilderInfo based on the given data
 func NewSlugBuilderInfo(appName string, shortSha string, disableCaching bool) *SlugBuilderInfo {
-	slugName := fmt.Sprintf("%s:git-%s", appName, shortSha)
-	tarKey := fmt.Sprintf("home/%s/tar", slugName)
+	basePath := fmt.Sprintf(GitKeyPattern, appName, shortSha)
+	tarKey := fmt.Sprintf("%s/tar", basePath)
```
bacongobbler: Was this intentionally changed? Mostly concerned about backwards compatibility. Looks harmless but just want to confirm.

pkg/gitreceive/slug_builder_info.go
```diff
@@ -19,12 +21,12 @@ type SlugBuilderInfo struct {
...
 	// this is where workflow tells slugrunner to download the slug from, so we have to tell slugbuilder to upload it to here
-	pushKey := fmt.Sprintf("home/%s/push", slugName)
+	pushKey := fmt.Sprintf("%s/push", basePath)
```
bacongobbler: Same here, is this an intentional change?

bacongobbler:
I think we just need someone to manually test that this doesn't break backwards compatibility and it's good to go. 👍

pkg/gitreceive/slug_builder_info.go
Joshua-Anderson: I just changed around patter so the tar key could be generated by the cleaner. That meant the slugName was no longer was the correct meaning... since it contained the entire path, so I renamed it.

mboersma:
I tested this change on top of Workflow v2.10.0 with S3 and an existing buildpack app, and also with multiple apps with multiple revisions to ensure it doesn't delete too much. Works as advertised. Thanks a ton @Joshua-Anderson!

PR #447, opened 2016-10-26
fix(slugbuilder): Dont expose the config as env vars during build

### LINKED ISSUE #444: Don't set env vars during compile phase
During the compile phase, heroku does not set user-specified environment variables/settings. Docs here: https://devcenter.heroku.com/articles/buildpack-api#bin-compile

In contrast, it appears that deis builder does set these variables. This is causing problems because we set PYTHONHOME and PYTHONPATH, which is causing builds to fail.

You should adhere to the heroku methodology of exposing settings in an envdir, but not actually setting them during the compile phase.

kmala:
fixes #444

### DIFF
--- a/pkg/gitreceive/build.go
+++ b/pkg/gitreceive/build.go
@@ -195,11 +195,21 @@ func build(
 		if !slugBuilderInfo.DisableCaching() {
 			cacheKey = slugBuilderInfo.CacheKey()
 		}
+		envSecretName := fmt.Sprintf("%s-build-env", appName)
+		err = createAppEnvConfigSecret(kubeClient.Secrets(conf.PodNamespace), envSecretName, appConf.Values)
+		if err != nil {
+			return fmt.Errorf("error creating/updating secret %s: (%s)", envSecretName, err)
+		}
+		defer func() {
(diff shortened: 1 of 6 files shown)


aboyett:
Looks good to me on the first pass. I'll try testing it tomorrow.

bacongobbler:
@kmala can you please describe what the PR's intention is here? There's a lack of documentation here on the implementation details here for reviewing.

kmala:
Heroku doesn't expose the app config as the environment vars and instead stores them in the env_dir which is passed as an arg to the compile https://devcenter.heroku.com/articles/buildpack-api#bin-compile.
I had created a secret with the config and mounted it in the env dir.

kmala:
I did a manual test by priniting the environemnt vars and the files in the /tmp/env dir we are passing to the compile.

bacongobbler:
Did you happen to test running multiple builds of different applications with this change as well? I'm concerned about concurrent builds tripping up each other, and that we handle proper cleanup of the secrets and all that if the build fails.

I'm sure the happy path works, but I would like to make sure that we handle the fail cases as well.

bacongobbler:
Looks like line 204 in build.go addresses my concern about proper cleanup and e2e ran this without issue, so code LGTM.

PR #489, opened 2017-02-22
fix(.travis.yml): declare env vars on one line

mboersma:
Without this change, Travis CI runs a separate [job for each environment variable](https://docs.travis-ci.com/user/environment-variables/#Defining-Multiple-Variables-per-Item), which was unintended.


kris-nova:
Oh there are lots of these 😃

Issue #410, opened 2016-08-18
nodeSelector to run dockerbuilder pod


arkkanoid:
Deploy an app consumes high CPU, in order to don't affect the deployed apps, I would like to run the dockerbuilder container in an isolate node. 
In the following method the builder component creates the dockerbuilder pod:

https://github.com/deis/builder/blob/96292153e5a52f8ea7643349665f8b3d75a5c1c8/pkg/gitreceive/k8s_util.go#L55

I could add 'nodeselector' parameter in the pod creation, but it means that I'll have a custom Deis component...
There is an esasier way to run this pod in an isolate node?

kmala:
i don't think there is another way than the one you suggested.If you want to make a PR against the builder instead of maintaining you own component, i am happy to help you. You can just add it to the struct https://github.com/deis/builder/blob/master/pkg/gitreceive/config.go#L15 and need to use it if specified while creating the pod.Any one who wan't to use can edit the deis-builder deployment object to add the extra environment variable.

kmala:
this can used to track as an enhancement request

PR #492, opened 2017-03-06
Add pod selector for builder job

zinuzoid:
add a new environment variable to inject builder job node selector.
format `BUILDER_POD_NODE_SELECTOR` = `network:fast,disk:ssd`

Closes https://github.com/deis/builder/issues/410
docs https://github.com/deis/workflow/pull/770

### DIFF
--- a/charts/builder/templates/builder-deployment.yaml
+++ b/charts/builder/templates/builder-deployment.yaml
@@ -94,6 +94,10 @@ spec:
                 secretKeyRef:
                   name: builder-key-auth
                   key: builder-key
+{{- if (.Values.builder_pod_node_selector) }}
+            - name: BUILDER_POD_NODE_SELECTOR
+              value: {{.Values.builder_pod_node_selector}}
+{{- end}}
           livenessProbe:
             httpGet:
               path: /healthz
--- a/charts/builder/values.yaml
+++ b/charts/builder/values.yaml
@@ -3,4 +3,4 @@ pull_policy: "Always"
 docker_tag: canary
 # limits_cpu: "100m"
 # limits_memory: "50Mi"
-# builder_pod_node_selector: "cloud.google.com/gke-nodepool=builder"
+# builder_pod_node_selector: "disk:ssd"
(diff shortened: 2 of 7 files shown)


deis-admin:
Thanks for the contribution! Please ensure your commits follow our [style guide](http://docs.deis.io/en/latest/contributing/standards/#commit-style-guide). This code will be tested once a Deis maintainer reviews it.

vdice:
Jenkins, test this please.

zinuzoid:
Jenkins, test this please.

vdice:
Jenkins, test this please.

charts/builder/templates/builder-deployment.yaml
```diff
@@ -94,6 +94,10 @@ spec:
                 secretKeyRef:
                   name: builder-key-auth
                   key: builder-key
+{{- if (.Values.builder_pod_node_selector) }}
```
vdice: Curious if there was a reason to go this route of `if` logic rather than something like:
```
- name: BUILDER_POD_NODE_SELECTOR
  value: {{ default "" .Values.builder_pod_node_selector}}
```
(would then be `BUILDER_POD_NODE_SELECTOR=` in builder env; perhaps not wanted)

vdice:
Closes https://github.com/deis/builder/issues/410

charts/builder/templates/builder-deployment.yaml
zinuzoid: not really a big reason for that, it would just be less variable in yml file. but can do either ways.

vdice:
Docs PR: https://github.com/deis/workflow/pull/770

mboersma:
I tested this manually and saw that both a dockerbuilder and slugbuilder pod will fail to schedule (sit in `Pending`) if a required label isn't met, and that applying that label satisfies the scheduler.

```shell
$ kubectl label nodes node1 disk=ssd
```
 Also poked around at the builder deployment and [docker|slug]builder pods to verify labels. Works as advertised--thanks a bunch @zinuzoid!

PR #495, opened 2017-03-14
feat(ingress) Experimental Native Ingress

kris-nova:
Adds LoadBalancer to builder if we are using `EXPERIMENTAL_NATIVE_INGRESS`

Requires deis/controller#1243
Requires deis/workflow#732
Requires https://github.com/deis/router/pull/316


charts/builder/templates/builder-service.yaml
```diff
@@ -11,3 +11,6 @@ spec:
       targetPort: 2223
   selector:
     app: deis-builder
+{{ if .Values.global.experimental_native_ingress }}
+  type: "LoadBalancer"
+{{ end }}
```
bacongobbler: needs newline character here

kris-nova:
@bacongobbler nit has been fixed 😄

kris-nova:
So this is all rebased and ready to go methinks..

The only thing it looks like we are missing is the fact that test coverage went down here. But I am not really sure what we would do in a situation like this as we are only really adding `YAML` configuration and not really any code here.

@mboersma thoughts?

mboersma:
> test coverage went down here

Probably just the vagaries of codecov--the YAML files aren't relevant obviously, and there isn't a unit test we could write to cover the `LoadBalancer` change anyway.

PR #498, opened 2017-03-20
ref(k8s_util): make users opt into DOCKER_BUILD_ARGS

bacongobbler:
injecting build args into the image by default has caused a significant performance penalty for
applications not utilizing the build args. Making it an opt-in feature flag via `deis config:set DEIS_DOCKER_BUILD_ARGS_ENABLED=1` allows users to inject their environment into their Dockerfiles, but they must opt into that performance penalty.

I feel like there's been enough concerns about turning this feature flag on by default in v2.12.0 that we should consider cutting a v2.12.1 with this fix.

### DIFF
--- a/pkg/gitreceive/k8s_util.go
+++ b/pkg/gitreceive/k8s_util.go
@@ -67,8 +67,10 @@ func dockerBuilderPod(
 	// {"KEY": "value"}
 	//
 	// So we need to translate the map into json.
-	dockerBuildArgs, _ := json.Marshal(env)
-	addEnvToPod(pod, "DOCKER_BUILD_ARGS", string(dockerBuildArgs))
+	if _, ok := env["DEIS_DOCKER_BUILD_ARGS_ENABLED"]; ok {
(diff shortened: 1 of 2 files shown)


chexxor:
I've heard some projects have hundreds of env vars, which sees significant build time increases. A project like that would never be able to opt-in to the "DOCKER_BUILD_ARGS" feature unless they choose exactly which subset of their env vars to include in the build.

So, another idea is to whitelist the env args to include in the build environment. This could like like `DEIS_DOCKER_BUILD_ARGS=ENV_VAR_1:ENV_VAR_2:ENV_VAR_3`. This could exist alongside the `DEIS_DOCKER_BUILD_ARGS_ENABLED` env var proposed in this patch, so my proposal doesn't need to burden the addition of this patch.

bacongobbler:
@chexxor this is sort of a stop-gap fix for now. Please feel free to implement that if you're feeling motivated to work on that, though!

PR #502, opened 2017-03-29
fix(gitreceive): fix checkForEnv logic

lshemesh:
val != val is clearly never going to fail. Changed
to expVal != val


deis-admin:
Thanks for the contribution! Please ensure your commits follow our [style guide](http://docs.deis.io/en/latest/contributing/standards/#commit-style-guide). This code will be tested once a Deis maintainer reviews it.

mboersma:
Jenkins, add to whitelist.

mboersma:
> codecov/project — 56.34% (-0.09%) compared to 4634fe1

I honestly can't see how that is the case, based on the diff in this PR.

